PR #2, opened 2017-12-04
Added search url.

blackbaud-brandonhare:


skyuxconfig.json
```diff
@@ -8,6 +8,7 @@
       "services": [
         {
           "title": "SKY UX 2",
+          "searchUrl": "https://stache-search-query.sky.blackbaud.com/search/skyux2",
```
Blackbaud-BobbyEarl: Should this change to `skyux` instead of `skyux2`?

blackbaud-brandonhare: Yes it should. I'll change that now.

PR #8, opened 2018-01-10
Added external flag

blackbaud-brandonhare:


Blackbaud-BobbyEarl:
Should we add a `skyuxconfig.serve.json` file to offset setting `host` to developer?

blackbaud-brandonhare:
Yeah, that's probably a good idea too.

PR #32, opened 2018-02-16
Fix tile event name, plus other doc tweaks.

blackbaud-johnly:
Based on https://github.com/blackbaud/skyux2/blob/master/src/modules/tiles/tile/tile.component.ts and https://github.com/blackbaud/skyux2/blob/master/src/modules/tiles/tile/fixtures/tile.component.fixture.html, I assumed we just need to change `isCollapsedChanged` to `isCollapsedChange`. While making this change I also noticed a misspelling ("bulding" instead of "building") and decided to make a few quick style/grammar fixes as well.

### DIFF
--- a/src/app/components/tile/index.html
+++ b/src/app/components/tile/index.html
@@ -26,7 +26,7 @@
     <sky-demo-page-property
       propertyName="isCollapsedChange"
     >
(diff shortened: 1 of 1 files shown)


blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/1481.

PR #34, opened 2018-02-21
Form error user assist guidelines

Blackbaud-ToddRoberts:


src/app/design/guidelines/form-design/index.html
```diff
@@ -0,0 +1,396 @@
...
+				<li><a routerLink="/components/autocomplete">Autocomplete</a></li>
+				<li><a routerLink="/components/checkbox">Checkbox</a></li>
+				<li>HTML <stache-code>&lt;select&gt;</stache-code> field</li>
+				<li>Lookup (component is not yet developed)</li>
+				<li><a routerLink="/components/radio">Radio button</a></li>
+				<li>SkyUX Select field (component is not yet developed)</li>
+			</ul>
+		</stache-column>
+		<stache-column screenMedium="10">
+			<div class="screenshot screenshot-info" style="min-width: 700px;">
```
blackbaud-johnly: In the flowchart:
- In the first column header, I'd change "there are to choose from" to "are in the list." 
- In the second column header, I'd change "the user" to "users."
- In the third column header, I'd change "the user" to "users" and "value" to "values" and "are looking for" to "want." ... Also, in the old version of these guidelines, we mentioned the need to search in the description for the HTML select field. Would it be any clearer what we mean by this column if we changed the label to "Do users need to search for specific values?" I'm not sure if that's more clear. Just wondering if it's closer to what we are getting at.
- In the fourth column header, I'd delete "this component" since one of the items is not a component.

In the fourth column:
- I'd remove the opening and closing tags around "select" and add "field." 
- I'd add "field" after "Autocomplete" and "Lookup."
- I'd change "SkyUX Select field" to "SKY UX select field."

PR #35, opened 2018-02-21
edit docs based on user feedback

Blackbaud-LindseyRix:
Received feedback from a user completing SKY UX prereqs for a new Stache site. This PR provides suggested edits based on the user's issues. 

- Didn't know how to download the SSL cert. (if you click the link, it just opens)
- Didn't realize when we say don't use node 7+ that means don't use numbers higher than 7 (ex. 8)
- The link to node was confusing too, since we take them to the latest release page.

I also re-worded the Git section to better match node.js section, and make it clearer you need to install Git.


blackbaud-johnly:
I made a few minor wording/grammar tweaks. Thanks for making these changes @Blackbaud-LindseyRix!

PR #39, opened 2018-02-27
Describe skyux watch in Basics > Run Tests

Blackbaud-TommyVernieri:
Previously `skyux watch` was mentioned, but not directly explained.


Blackbaud-TommyVernieri:
Thanks @blackbaud-johnly I'm happy with that change.

PR #40, opened 2018-02-28
Close stache-code block

jeffbdye:
Quick fix from glancing over the localization docs


jeffbdye:
Perhaps just need @blackbaud-johnly's feedback on this since nobody was assigned to the PR automatically

PR #41, opened 2018-02-28
Call out that builder converts variable to string.

blackbaud-johnly:
Updated the route parameters docs based on feedback this morning.

### DIFF
--- a/src/app/learn/get-started/advanced/route-params/index.html
+++ b/src/app/learn/get-started/advanced/route-params/index.html
@@ -7,7 +7,7 @@
     In addition to automatically registering routes and creating components for all <stache-code>index.html</stache-code> files in the <stache-code>src/app</stache-code> directory, SKY UX supports routes that contain <a href="https://angular.io/guide/router#route-definition-with-a-parameter">route parameters</a>.
   </p>
   <p>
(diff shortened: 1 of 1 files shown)


src/app/learn/get-started/advanced/route-params/index.html
```diff
@@ -7,7 +7,7 @@
...
   </p>
   <p>
-    To create a route parameter, you create a subfolder and prefix the name with an underscore. The string after the underscore is used as the variable name, so it must be a valid TypeScript variable. For example, to use a route parameter in a <stache-code>src/app/dogs</stache-code> folder, you can create a <stache-code>src/app/dogs/_id</stache-code> subfolder and then create an <stache-code>index.html</stache-code> file with the following content:
+    To create a route parameter, you create a subfolder and prefix the name with an underscore. The subfolder name must be a valid TypeScript variable because SKY UX Builder converts the portion after the underscore to a string and uses it as the variable name. For example, to use a route parameter in a <stache-code>src/app/dogs</stache-code> folder, you can create a <stache-code>src/app/dogs/_id</stache-code> subfolder and then create an <stache-code>index.html</stache-code> file with the following content:
```
Blackbaud-BobbyEarl: I think we need to tweak this to put emphasis that the value (or type) of the variable created is a string.

blackbaud-johnly: How about changing `The subfolder name must be a valid TypeScript variable because SKY UX Builder converts the portion after the underscore to a string and uses it as the variable name.` to `The subfolder name must be a valid TypeScript variable because SKY UX Builder creates a variable with that name (minus the underscore). The value of the variable is a string.`?

Blackbaud-BobbyEarl: I like that, but maybe "type" instead of "value"?

blackbaud-johnly: Done.

PR #44, opened 2018-02-28
Adding missing last step in tutorial

Blackbaud-MikeBentley:
Adding the fact that the index.html file needs to be updated to reference the component as well as pass in the id input value.

### DIFF
--- a/src/app/learn/get-started/advanced/route-params/index.html
+++ b/src/app/learn/get-started/advanced/route-params/index.html
@@ -201,6 +201,14 @@ export class AboutUserComponent implements OnInit {
     </li>
     <li>
       <p>
+        Update index.html to use the about user component selector
+      </p>
+      <stache-code-block>
+        <my-about-user [userid]="id"></my-about-user>
+      </stache-code-block>
+    </li>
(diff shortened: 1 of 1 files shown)


src/app/learn/get-started/advanced/route-params/index.html
```diff
@@ -199,6 +199,14 @@ <h1>About our Team</h1>
 </div>
       </stache-code-block>
     </li>
+    <li>
+      <p>
+        Update index.html to use the about user component selector
```
blackbaud-johnly: In the current step 8, we say "add index.html to create a page and insert the my-about-user component." and have the same code snippet in place within a longer code sample. It looks like I need to capitalize "I" in "userId" in that code sample, but I think the rest of this new step just repeats step 8.

blackbaud-johnly:
Went over this with @Blackbaud-BobbyEarl, and he agreed that it looks like the added step is already covered in the existing step 8. I'm closing the PR, but please just let me know if I'm missing somethign @Blackbaud-MikeBentley. Thanks.

PR #50, opened 2018-03-05
[HOLD] Status indicator guidelines

Blackbaud-ToddRoberts:
This needs to be held until the corresponding changes to labels and status text make it into a SkyUX build.


src/app/components/status-indicator/index.html
```diff
@@ -0,0 +1,112 @@
...
+	The status indicator classes create styled text to draw attention to data that needs to be addressed. A status is "a state or condition at a particular time". Statuses fall into four categories: success, warning, danger, and info. Examples include late, submitted, awaiting approval, etc.
+</p>
+
+<stache-page-anchor>
+	Status indicator classes
+</stache-page-anchor>
+
+  <div class="definition-table definition-table-wide">
+    <div class="definition-item">
+      <div class="sky-text-info">Info</div>
```
blackbaud-johnly: When I serve this locally, there is no icon before "Info."

Blackbaud-ToddRoberts: It won't show up until that change is merged/released in SKY UX

src/app/components/status-indicator/index.html
```diff
@@ -0,0 +1,112 @@
...
+  </div>
+
+  <div class="definition-table definition-table-wide">
+    <div class="definition-item">
+      <div class="sky-text-warning">Warning</div>
+    </div>
+    <div class="definition-props">
+      <div class="definition-usage">
+        <div class="sky-field-label">Usage</div>
+        <div>This class prepends a yellow warning icon to provide a compact way to indicate a "warning" status.</div>
```
blackbaud-johnly: We say "yellow," but it looks orange to me. We also say "yellow" at https://developer.blackbaud.com/skyux/design/styles/typography, but is that correct?

src/app/components/status-indicator/index.html
```diff
@@ -0,0 +1,112 @@
...
+<stache-page-anchor>
+	UX guidelines
+</stache-page-anchor>
+	
+	<div class="sky-subsection-heading">Status importance</div>
+		<p>Some statuses are crucial to a user's task, while others are no more important than other information on the page or page area. If they are crucial to a user's task you should use a status indicator. If they are ancillary, treat them as regular text.</p>
+		<p>Just because one status in an area is crucial, does not mean they all are. For example, a list of invoices could show a mix of paid, unpaid and overdue invoices. It's likely that in that case only the overdue invoices would need a status indicator if they are the only invoices that need immediate attention and action.</p>
+			
+	<div class="sky-subsection-heading">Status indicator placement</div>
+		<p>Status indicators should be used in an inline context tied to a specific piece of data, rather than in a summary area. For example, an indicator can be placed in a grid cell or card. If you need to convey a higher-level status in a summary area, use a <a routerLink="label">label</a> instead.</p>
```
blackbaud-johnly: I think this warrants a guidelines topic for status information. I think it would be useful to explicitly explain when to use a status indicator vs. a label vs. an alert vs. whatever other similar components we have.

blackbaud-johnly: So, looking at the call out info guidelines, I guess we have that covered already.  Although it strikes me as possibly confusing that alerts convey status but we don't mention them in the Status section.

Blackbaud-ToddRoberts: We are going to be revisiting the "Call out info" info guidelines to tie the other components referenced on the page together better, but want to get these up in the interim as an incremental improvement.

blackbaud-johnly: Sounds good. Thanks.

src/app/components/status-indicator/index.html
Blackbaud-ToddRoberts: It's pretty much halfway between yellow and orange, but we went with the yellow description to reflect the stoplight metaphor it's based on.

blackbaud-johnly: OK. Thanks.

src/app/design/guidelines/call-out-info/index.html
```diff
@@ -28,10 +28,23 @@
...
+		<p>Just because one status in an area is crucial, does not mean they all are. For example, a list of invoices could show a mix of paid, unpaid and overdue invoices. It's likely that in that case only the overdue invoices would need a status indicator if they are the only invoices that need immediate attention and action.</p>
+			
+	<div class="sky-subsection-heading">Placement</div>
+		<ul>
+			<li>You should use a <a routerLink="/components/label">label</a> in a summary context. For example, you can place a label in a page summary.</li> 
+			<li>You should use a <a routerLink="/components/status-indicator">status indicator</a> in an inline context tied to a specific piece of data. For example, you can place an indicator in a grid cell.</li>
+		</ul>
+	
+	<div class="sky-subsection-heading">Status presentation decision tree</div>
+		<div class="screenshot screenshot-info"><img src="~/assets/img/guidelines/calloutinfo/status-decision-tree.png" style="width: 100%;" /></div>
```
blackbaud-johnly: I have a few edits to suggest for the decision tree:

In the left-hand column, I'd change "this" to "the status" and "show up" to "appear" in the first item. In the second item, I'd change "user's task(s)" to "user tasks." And I'd delete "this" in the third item (or just delete the third item altogether).

In the middle column, why do we have "(summary)" in the bottom left? Seems like it should just be "Label." ... And in the bottom right, what do we mean by "Sky Typography"? Do we mean "Regular text"? Saying "Sky Typography" sounds like we are referencing a specific tool that we haven't referenced, so I'd change it to "Regular text." (If not, "SKY" should be all uppercase and "typography" should be lowercase.

In the right column, why do we have "(inline)" in the bottom left? Seems like it should just be "Status indicator." ... And in the bottom right, same question/suggestion as in the middle column.

Blackbaud-ToddRoberts:
I updated the decision tree if you can take a look at that

PR #64, opened 2018-03-27
added documentation for the infinite-scroll component

blackbaud-conorwright:
Relies on: https://github.com/blackbaud/skyux2/pull/1574/files


src/app/components/infinite-scroll/index.html
```diff
@@ -0,0 +1,29 @@
...
+    The infinite scroll component allows users to load data as the page scrolls.
+  </sky-demo-page-summary>
+
+  <sky-demo-page-properties>
+    <sky-demo-page-property
+      propertyName="hasMore"
+      defaultValue="false"
+      isOptional="true"
+    >
+    A boolean specifying whether there is more data to load. If this is false, infinite scroll will hide and stop calling the load function.
```
blackbaud-johnly: I'd change the first sentence to "Indicates whether more data is available to load."

I'd change "this is" to "set to" and format "false" with "stache-code" tags. I'd also change "will hide and stop" to "hides and stops."

src/app/components/infinite-scroll/index.html
```diff
@@ -0,0 +1,29 @@
...
+    </sky-demo-page-property>
+
```
blackbaud-johnly: Insert a closing  "sky-demo-page-properties" tag here.

src/app/components/infinite-scroll/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <sky-demo-page-property
```
blackbaud-johnly: Insert an opening "sky-demo-page-properties" here with "sectionHeading="Events" " to create a separate section with an Events header.

src/app/components/infinite-scroll/index.html
```diff
@@ -0,0 +1,29 @@
...
+      propertyName="onLoad"
+      isOptional="false"
+    >
+    This is an event that fires when when the load more button is clicked or a scroll triggers a need to load more data. This will only emit an event when the hasMore property is `true` and data is not currently loading.
```
blackbaud-johnly: Delete "This is an event that" and just start with "Fires" for consistency with other component properties/events. Also, delete the extra "when" and change "the load more button is clicked" to "users click the load more button" to use active voice. ... And after "triggers," I'd change "a" to "the."

Change "will only emit" to "only emits." ... Wrap "hasMore" in "stache-code" tags and change the tick marks around "true" to "stache-code" tags. ... I think we can delete "currently" without changing the meaning.

PR #65, opened 2018-03-27
Document SkyAppAssetsService and update existing static assets docs.

blackbaud-johnly:


blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/996.

Blackbaud-BobbyEarl:
Hey @blackbaud-johnly, I'm guessing there's a problem if we don't encode the `{` as `&#123;`?

blackbaud-johnly:
@Blackbaud-BobbyEarl, I added `languageType` to `stache-code-block` so that we don't need to escape `{`. ... I still get errors if I don't escape `~` though, so I left that as is.

PR #67, opened 2018-03-29
Skyux module info

Blackbaud-PaulCrowder:
https://github.com/blackbaud/skyux2/issues/1588

### DIFF
--- a/src/app/components/shared/demo-page.component.html
+++ b/src/app/components/shared/demo-page.component.html
@@ -4,7 +4,7 @@
 
   <stache-page-summary>
     <h4 *ngIf="skyuxModulesForDisplay">
-      Defined in
+      <span class="sky-field-label">Defined in</span>
       <span *ngFor="let moduleName of skyuxModulesForDisplay">
         <stache-code>{{ moduleName }}</stache-code>
       </span>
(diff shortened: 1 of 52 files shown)


src/app/components/text-expand/index.html
```diff
@@ -1,5 +1,7 @@
 <sky-demo-page
-  pageTitle="Text expand">
+  pageTitle="Text expand"
+  [skyuxModule]="['SkyTextExpandModule', 'SkyTextExpandRepeaterModule']"
```
blackbaud-johnly: Should we add a comma or "and" between modules when we have more than one?

Blackbaud-PaulCrowder: I figured since they're stylized with the `<stache-code>` element and separated by a space that we didn't need any kind of grammatical conjunction between them.

src/app/components/shared/demo-page.component.html
```diff
@@ -3,6 +3,12 @@
   [pageTitle]="pageTitle">
 
   <stache-page-summary>
+    <h4 *ngIf="skyuxModulesForDisplay">
+      Defined in
```
blackbaud-johnly: Does this need the `sky-field-label` style like Todd suggested in Slack?

PR #70, opened 2018-04-03
added sketch kit

Blackbaud-ToddRoberts:


src/app/design/downloads/index.html
```diff
@@ -0,0 +1,12 @@
+<stache 
+  layout="container"
+  pageTitle="Downloads"
```
blackbaud-johnly: What do you think of "Templates" as a title? I assume the emphasis in the title and navigation should be that the section provides templates, not that it includes something downloadable.

src/app/design/downloads/index.html
```diff
@@ -0,0 +1,12 @@
...
+  windowTitle="Downloads">
+  
+<h2 class="sky-section-heading">Sketch UI kit</h2>
+<p>A Sketch file of the SKY UX components. Updated monthly, so check back from time to time to ensure you have the latest and greatest.</p>
+<div>
+    <a href="~/assets/design/skyux-kit.zip">Download the Sketch UI kit</a>
```
blackbaud-johnly: Do you think it would be useful to make the link standout by making it a button and wrapping it in the "sky-btn sky-btn-default" class?

PR #93, opened 2018-05-29
Remove stache-code for users who copy/paste

jeffbdye:
Certain folks using the docs may not have Stache2 installed in their demo SPAs, and so snippets containing components from it will throw unexpected errors. There's probably a better way to highlight those two words (id and MyAboutUser), but just wanted to call it out as something worth fixing due to this thread here: https://blackbaud.slack.com/archives/C3TQ4C0KG/p1527619236000336


blackbaud-johnly:
Thanks @jeffbdye. I addressed this in https://github.com/blackbaud/skyux2-docs/pull/94. I also double-checked the rest of the docs to make sure this copy/paste error didn't occur anywhere else.

PR #97, opened 2018-05-31
sort guidelines

Blackbaud-ToddRoberts:


src/app/components/list/toolbar/index.html
```diff
@@ -126,5 +126,16 @@ <h3>
     <sky-list-toolbar-custom-demo></sky-list-toolbar-custom-demo>
     <sky-demo-page-code demoName="List toolbar"></sky-demo-page-code>
   </sky-demo-page-example>
+	
+	<sky-demo-page-content sectionHeading="UX guidelines">
+		<div class="sky-subsection-heading">Sorting</div>
+		<p>If you use the toolbar with a grid view, you have two options for letting users sort the list:</p>
+		<ul>
+			<li>The sort button in the toolbar</li>
+			<li>Column header click sorting</li>
```
blackbaud-johnly: This seems like awkward wording to me. I get what we mean, but I feel like we're missing hyphens to indicate what modifies what. But I'm not sure how to hyphenate this to make it clearer. Also, "sorting" seems somewhat redundant. What do you think of changing this to "Clickable column headers"?

PR #131, opened 2018-08-14
[HOLD] Using headings instead of a list for configuration options.

Blackbaud-BobbyEarl:
I was curious to see what this would look like @blackbaud-johnly


Blackbaud-DeniseCisneros:
Personally, I like this switch because it provides better guidance and having the nav on the right will be a little easier on the eyes.

blackbaud-johnly:
Closing in favor of https://github.com/blackbaud/skyux2-docs/pull/233.

PR #135, opened 2018-08-16
Document selectedColumnIdsChange property for list view grids

blackbaud-johnly:
Resolves https://github.com/blackbaud/skyux2/issues/1906. ... The update is merged in https://github.com/blackbaud/skyux2/pull/1897, but marking as hold until the next SKY UX release.


Blackbaud-TrevorBurch:
These look good. 

The event does also fire when you reorder the columns and always fires with the array contents in the same order as they are displayed. Should we call that out at all?

blackbaud-johnly:
I think we should call out that info. And since the VSTS check failed anyway, I'll go ahead and make the change get that in the docs and to also kick the checks off again.

PR #92, opened 2018-05-29
added radio group description to radio docs

blackbaud-conorwright:
Dependent on PR: https://github.com/blackbaud/skyux2/pull/1692

PR #153, opened 2018-09-07
added radio group description to radio docs (#92)

Blackbaud-BobbyEarl:
Originally from: https://github.com/blackbaud/skyux2-docs/pull/92


Blackbaud-BobbyEarl:
I don't think this is being rendered as intended.  
screen shot 2018-09-07 at 9 50 39 am

blackbaud-johnly:
FYI that I am looking into this now as it is tied to https://github.com/blackbaud/skyux2/issues/1942.

blackbaud-johnly:
@blackbaud-conorwright, I think I updated the docs to reflect all the points we discussed, but please let me know if you see any problems or have any suggestions. Thanks!

src/app/components/radio/index.html
```diff
@@ -4,57 +4,103 @@
...
+      propertyName="ariaLabelledBy"
+      isOptional="true"
     >
-      Specifies ...
+      Sets the radio button group's <stache-code>aria-labelledby</stache-code> attribute to support accessibility. The value should be the HTML element ID (without the leading #) of the element that labels the radio button group. This property accepts <stache-code>string</stache-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="name"
+      isOptional="true">
+      Specifies a name for the collection of radio buttons that the component groups together. If a <stache-code>name</stache-code> property is also set on the <stache-code>sky-radio</stache-code> elements, then this name overwrites it. If a <stache-code>name</stache-code> property is not set on the <stache-code>sky-radio</stache-code> elements, then this property is required.
```
blackbaud-conorwright: Sorry, I know we talked about this, but I've been thinking about it and do you think it might be better to list this as required (isOptional=false) and list the radio group and it's properties first on the page? So that this is the first thing they see? If not then this is fine :)

I've also been wondering about the ordering of properties in general and despite knowing what happened the last time I asked about the ordering of properties in documentation 😜 , I kind of have to know: Have we considered listing required properties first and then optional ones? It seems like it would be more straightforward to look and see what I must have on the component to use it.

blackbaud-johnly: Sorry for the slow response on this. I agree with both of your suggestions, and I followed up with Steve who also agreed. So I removed the "optional" label from the name property, moved the group properties to the top of the page, and reorganized the properties so that required properties appear in alphabetical order followed by optional properties in alphabetical order. Going forward, I'll make that same change to properties in other component docs as well.

Thanks for the feedback. ... With those changes in place, does this look OK to merge?

PR #167, opened 2018-10-01
added options for styleType

Blackbaud-DeniseCisneros:
Added `primary` and `list` but not `secondary` because that's what the default value looks like but can edit to be more specific. 

This addresses [Issue #1866](https://github.com/blackbaud/skyux2/issues/1866)

### DIFF
--- a/src/app/components/confirm/index.html
+++ b/src/app/components/confirm/index.html
@@ -77,7 +77,7 @@
       propertyName="styleType"
       isOptional="true"
     >
-    A string that describes the button's visual appearance. <em>(Default: "default")</em>.
+    A string that describes the button's visual appearance. This property also accepts the <stache-code>primary</stache-code> and <stache-code>link</stache-code> values listed for <a stacheRouterLink="/components/button" fragment="classes">Button classes</a>. <em>(Default: "default")</em>.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="action"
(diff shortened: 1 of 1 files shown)


src/app/components/confirm/index.html
```diff
@@ -77,7 +77,7 @@
       propertyName="styleType"
       isOptional="true"
     >
-    A string that describes the button's visual appearance. <em>(Default: "default")</em>.
+    A string that describes the button's visual appearance. This property also accepts the <stache-code>primary</stache-code> and <stache-code>link</stache-code> values listed for <a stacheRouterLink="/components/button" fragment="classes">Button classes</a>. <em>(Default: "default")</em>.
```
blackbaud-johnly: "also accepts" strikes me as weird since the default appears at the end. It's like we're alluding to something we haven't mentioned yet. Also, after looking at it a little more, I think "describes the button's visual appearance" is sorta vague when we can say that we are applying those SKY UX styles. What do you think of rewording the entire entry a bit to something like:
"styleType — (Optional.) A string that specifies a SKY UX button class to apply to the button. The valid options are `primary`, which formats the primary action on the dialog, `default`, which formats a secondary action, and `link`, which formats an action that takes users to another page."    (Note: I'm thinking "a SKY UX button class" would be the link to the button classes.)

Also, we don't include the default in the description. We set it through a property above. So you can add `defaultValue="default"` alongside the `propertyName` and isOptional` values in the lines above the description.

PR #171, opened 2018-10-11
added ariaLabel to properties

Blackbaud-DeniseCisneros:
John, feel free to English me on that line, and Steve, keep me honest, please. :)

### DIFF
--- a/src/app/components/wait/index.html
+++ b/src/app/components/wait/index.html
@@ -33,7 +33,7 @@
       propertyName="ariaLabel"
       isOptional="true"
     >
-    Defines a string value to label the wait component and the element or page being waited on for accessibility.
+    Defines a string value to label the wait icon for accessibility while an element or page loads.
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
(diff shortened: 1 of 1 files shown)


src/app/components/wait/index.html
```diff
@@ -29,6 +29,12 @@
     >
     When set to <stache-code>true</stache-code>, wait indication appears on the page level instead of the parent element level. We recommend that you use the <stache-code>beginBlockingPageWait</stache-code> or <stache-code>beginNonBlockingPageWait</stache-code> functions of the <stache-code>SkyWaitService</stache-code> instead of setting this on the component level.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="ariaLabel"
+      isOptional="true"
+    >
+    Defines a string value to label the wait component and the element or page being waited on for accessibility.
```
blackbaud-johnly: I don't think this labels "the element or page being waited on." I think it just labels the spinning wait icon so that users with disabilities know that the page is loading. But I'd defer to @Blackbaud-SteveBrush to confirm. ... If I'm right, I think we should tweak this just a bit to be something like: 

"Defines a string value to label the wait icon for accessibility while an element or page loads."

Is that right, @Blackbaud-SteveBrush?

Blackbaud-SteveBrush: That's correct, @blackbaud-johnly . The ARIA attributes are being applied to the wait component directly.

Blackbaud-DeniseCisneros: Thanks, Steve! I'll edit that line to go with John's wording.

PR #175, opened 2018-11-02
Added library as an option for skyux new

Blackbaud-BrandonJones:


src/stache/includes/cli/template.html
```diff
@@ -7,7 +7,10 @@
   <li>
     <stache-code>-t stache</stache-code> — References the <a href="https://github.com/blackbaud/skyux-template-stache">Stache template</a> to create <a href="https://developer.blackbaud.com/stache/learn/get-started"><strong>public</strong> Stache sites</a> for technical documentation.
   </li>
+  <li>
+    <stache-code>-t library</stache-code> - References the <a href="https://github.com/blackbaud/skyux-template-library">SKY UX Library template</a> to create <a href="https://developer.blackbaud.com/skyux/learn/reference/component-libraries">SKY Libraries</a> that allow developers to build new reusable front-end libraries can be published for other SPAs to consume.
```
blackbaud-johnly: I'd put this new entry before the `-t stache` entry instead of after it to stick with alphabetical order.

The hyphen between the code snippet and the description should be a long dash. (You can copy and past from one of the other entries.)

Lowercase the "L" in "SKY UX library template." ... Lowercase the "L" and add "UX" to "SKY UX libraries." ... And I'd change "to build new reusable front-end libraries can be published for other SPAs to consume" to "to build and publish reusable front-end libraries for other SPAs to consume." (If you don't like the rewrite, we at least need to change "can be published " to "and publish them.)"

Blackbaud-BrandonJones: Done :)

PR #179, opened 2018-12-04
Document tile help button property and event

blackbaud-johnly:


src/app/components/tile/index.html
```diff
@@ -27,6 +34,11 @@
   </sky-demo-page-properties>
 
   <sky-demo-page-properties sectionHeading="Tile events">
+    <sky-demo-page-property
+      propertyName="helpClick"
```
Blackbaud-TrevorBurch: We probably want to call out that you have to have both `showHelp` set to true and be listening to `helpClick` in order for the help icon to appear. 

Settings works the same way.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/179.

PR #182, opened 2018-12-06
Summary action bar

blackbaud-johnly:


Blackbaud-MattGregg:
@Blackbaud-ToddRoberts I think the Anatomy section would be more useful ahead of the behavior section. The behavior section seems to reference or assume someone already knows the anatomical parts.

This sentence in the first usage section seems incomplete or fragment "Actions that can be taken in bulk on selected items in a multiselect." Wondering if this can be clearer. Something like Actions can be taken on multiple items in a list of items.

Also wondering why the split panel is listed as related but isn't linked and included in the usage section given it doesn't really exist in SKY UX yet and we haven't moved to new content strategy?

Can this be used in the Flyout if someone is using a list in there? Thinking yes, but could be good to call this out. Related, not advised to use in tiles?

Blackbaud-ToddRoberts:
@Blackbaud-MattGregg made updates based on your feedback. Regarding flyout, to me it's less of whether they are used in a flyout vs whether what's in the flyout would appropriately use a summary action bar. The main (only?) case would be if there is a multiselect list in the flyout, in which case it would be appropriate. We could call this out, but there is then the question of what needs to explicitly be called out as usable in flyouts when the answer is basically everything since a flyout is a page view container.

Blackbaud-AdamFunderburk:
Looks really good. I just had two points:

Should "List view grid" be added as another related component?
Can we add an negative example to the point about not using Summary action bar inside tiles or page sections? It's the only point not accompanied by an example and leaves an abrupt transition into the next section of the document.

Blackbaud-ToddRoberts:
@Blackbaud-AdamFunderburk added link to list view grid.

I thought about what to show in the "don't use in a tile" section - what needs to be illustrated is that items are selected but the action bar is off the page, which I haven't figured out a good way to illustrate.

src/app/components/summary-actionbar/index.html
```diff
@@ -0,0 +1,237 @@
+<sky-demo-page
+  pageTitle="Summary action bar"
+  skyuxModule="SkySummaryActionBarModule"
+>
+
+  <sky-demo-page-summary>
+    <p>
+      The summary action bar component provides a docked container for actions and summary information related on a form or multiselect list. The <stache-code>sky-summary-action-bar</stache-code> element implements the summary action bar, and within this element, the <stache-code>sky-summary-action-bar-actions</stache-code> and <stache-code>sky-summary-action-bar-summary</stache-code> elements contain the actions and sumamry information respectively.
```
Blackbaud-TrevorBurch: Should we call out the individual action types here? Or is that overkill? We have them listed seperately below but don't mention them in this top block.

Blackbaud-TrevorBurch: We also don't mention the summary element and so the only place it is currently mentioned is the usage guidelines. We should probably mention that.

.nvmrc
```diff
@@ -1 +1 @@
-6
+lts/carbon
```
Blackbaud-SteveBrush: Try `lts/dubnium`?

Blackbaud-TrevorBurch: Done

PR #183, opened 2018-12-06
Grid guidelines

blackbaud-johnly:


src/app/components/grid/index.html
```diff
@@ -128,4 +170,227 @@
...
+				<img src="~/assets/img/guidelines/grid/grid-reorder.png" style="width: 100%;" />
+			</div>
+		</stache-column>
+	</stache-row>
+    
+	<header><div class="sky-subsection-heading">Resizing columns</div></header>
+
+  <stache-row>
+    <stache-column screenSmall="6">
+			<p>The user can resize columns via drag and drop or the keyboard.</p>
```
Blackbaud-AlexKingman: "Drag and drop" seems fitting when talking about reordering columns. However, is that the right term when referring to resizing? Would something like "drag to resize" be more fitting?

src/app/components/grid/index.html
```diff
@@ -8,49 +8,91 @@
...
+    Specifies the data for the grid. Each item requires an ID and a property that maps to the <stache-code>field</stache-code> or <stache-code>id</stache-code> property of each column in the grid. This property accepts an array of <stache-code>any</stache-code> values.
     </sky-demo-page-property>
     <sky-demo-page-property
-      propertyName="selectedColumnIds"
+      propertyName="enableMultiselect"
+      defaultValue="false"
       isOptional="true"
     >
-    Specifies an array of columns to display in the grid. Will match the <stache-code>id</stache-code> or <stache-code>field</stache-code> properties of the columns. If no array is provided, then the grid displays all columns. Acceptable values: <stache-code>Array of strings</stache-code>.
+    Indicates whether to enable the multiselect feature that displays a column of checkboxes on the left side of the grid. To display the checkboxes, you must also specify the <stache-code>multiselectRowId</stache-code> property. This property accepts <stache-code>boolean</stache-code> values.
```
Blackbaud-AlexKingman: `To display the checkboxes, you must also specify the <stache-code>multiselectRowId</stache-code> property.`
This is no longer true. If you don't provide the `multiselectRowId` property, then it defaults to the `id` property of the provided data.

src/app/components/grid/index.html
blackbaud-johnly: I agree that this seems like the wrong term. I think this should be something like "The user can resize columns by clicking and dragging or by using the keyboard." Or maybe "The user can resize columns using the mouse or the keyboard." ... @Blackbaud-ToddRoberts, what do you think?

Blackbaud-ToddRoberts: @blackbaud-johnly your first option sounds good to me

src/app/components/grid/index.html
```diff
@@ -128,4 +170,227 @@
...
+				<header><div class="sky-section-heading">Usage</div></header>
+				<p>Grid provides a table-based layout.</p>
+			</stache-column>
+			<stache-column screenSmall="4" screenMedium="3">
+				<header><div class="sky-section-heading">Related components</div></header>
+				<ul>
+					<li><a routerLink="/components/list/grid">List view grid</a></li>
+					<li><a routerLink="/components/card">Card</a></li>
+					<li><a routerLink="/components/repeater">Repeater</a></li>
+					<li><a routerLink="/components/summary-actionbar">Summary action bar</a></li>
```
Blackbaud-ToddRoberts: Will need to remove this until summary action bar is in

Blackbaud-ToddRoberts:
preview link https://host.nxt.blackbaud.com/skyux?_sp.spa=pr-grid-guidelines

Blackbaud-AlexKingman:
@blackbaud-johnly Per the suggestions from the grids demo PR /blackbaud/skyux2#2230 we might need to offer guidance about when and how to use the select all/clear all toggles, and the "only show selected items" checkbox for the multiselect feature. We are going to add this toolbar to `list-view-grids`, but that work has not yet been done. How would you like to proceed here?

blackbaud-johnly:
@Blackbaud-AlexKingman, I looked at the note from the demo again, and while your note here referred to future work, the note there sounds like the note is not dependent on future work. If it's the former case, I still think we should create a docs issue as I suggested above. But if it's the latter case, what do you think about adding a sentence to the end of the `enableMultiselect` property like "If you need to include options to select and clear all the multiselect checkboxes, we recommend using the list view grid component." ... Let me know what you think.

Blackbaud-ToddRoberts:
@Blackbaud-MattGregg updated UX guidelines based on your feedback.

Blackbaud-AlexKingman:
@blackbaud-johnly Yep, I like what you suggested. Something like that -->
`If you need to include options to select and clear all multiselect checkboxes, we recommend using the list view grid component.`
The work on list view grid is actually beginning today, so there won't be that much gap between documenting this and actually having the ability to do it :)

src/app/components/grid/index.html
```diff
@@ -4,53 +4,116 @@
...
+    <sky-demo-page-property
+      propertyName="columns"
+    >
+    Specifies columns and column properties for the grid. This property accepts an array of <stache-code>SkyGridColumnModel</stache-code> values.
+    </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="data"
     >
-    Specifies an array of data for the grid. Each item in the array should have an ID and a property that maps to either the <stache-code>field</stache-code> or <stache-code>id</stache-code> property of each column in the grid.
+    Specifies the data for the grid. Each item requires an ID and a property that maps to the <stache-code>field</stache-code> or <stache-code>id</stache-code> property of each column in the grid. This property accepts an array of <stache-code>any</stache-code> values.
```
Blackbaud-AlexKingman: > Each item requires an ID and a property...
Should this instead be:
> Each item requires an `id` and a property...
?

Blackbaud-AlexKingman: Also, I always felt that the structure of `data` was pretty unclear until you scroll all the way down and look at the example below. Should we show a short example here? Or is that against our typical pattern?

src/app/components/grid/index.html
```diff
@@ -103,29 +160,281 @@
...
+						<span class="sky-deemphasized">Toolbar (optional)</span>
+					</li>
+					<li>
+						Header row
+					</li>
+					<li>
+						<span class="sky-deemphasized">Row selection checkbox (optional)</span>
+					</li>
+					<li>
+						<span class="sky-deemphasized">Context menu button (optional)</span>
```
Blackbaud-AlexKingman: I'm curious - "List count" and "Context menu button" aren't native functionality of the grid. The consumer will have to build them. Listing them here makes it seem like they're offered in this library. Is that ok? Or maybe this is more of a guideline of how to use those types of things if you want to build them yourself?

src/app/components/grid/index.html
```diff
@@ -103,29 +160,281 @@
...
+		</stache-column>
+		<stache-column screenSmall="6">
+			<div class="screenshot screenshot-success">
+				<img src="~/assets/img/guidelines/grid/grid-resize.png" style="width: 100%;" />
+			</div>
+		</stache-column>
+	</stache-row>
+    
+	<header><div class="sky-section-heading">Responsiveness</div></header>
+  <p>Grids tend to perform poorly in small viewports. The columns in fixed width grids will shrink down to the point of being unreadable. Scrolling grids will require a lot of interaction to see all the content.</p>
```
Blackbaud-AlexKingman: This paragraph is nice, but I'm left wondering what is the alternative? Is there a solution or suggested path to take here if I want to display data in small viewports?

This might not be a question we have to answer today, but thought it was worth mentioning that this is the last thing on the page, and it kind of leaves me hanging :)

Blackbaud-ToddRoberts: added a link to use repeater instead.

src/app/components/grid/index.html
blackbaud-johnly: I assume it should be `id`. I'll make the change.

As for a short example, I think it would be great. What should the example look like?

src/app/components/grid/index.html
Blackbaud-AlexKingman: Makes sense. That's kind of what I figured, but just wanted to mention it. I don't think it should hold up the PR. Just something to think about as we flesh these docs out.

PR #185, opened 2018-12-10
Add note linking to SKY UX 1

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/2224. ... @Blackbaud-BenLambert, do you think this is addresses the issue sufficiently?

### DIFF
--- a/src/app/learn/overview/index.html
+++ b/src/app/learn/overview/index.html
@@ -20,6 +20,9 @@
       SKY UX enables developers to focus on features instead of infrastructure. It creates user experience interactions and allows feature engineers to consume them.
     </p>
   -->
+  <sky-alert alertType="info">
+    This site describes the version of SKY UX that supports versions of Angular after the AngularJS (1.x) framework. For information about the version of SKY UX that supported the AngularJS (1.x) framework, visit <a href="https://skyux.developer.blackbaud.com/">skyux.developer.blackbaud.com/</a>.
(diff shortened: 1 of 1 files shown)


Blackbaud-BenLambert:
Yes, thanks!

src/app/learn/overview/index.html
```diff
@@ -20,6 +20,9 @@
       SKY UX enables developers to focus on features instead of infrastructure. It creates user experience interactions and allows feature engineers to consume them.
     </p>
   -->
+  <sky-alert alertType="info">
+    This site describes the version of SKY UX that supports versions of Angular after the AngularJS (1.x) framework. For information about the version of SKY UX that supported the AngularJS (1.x) framework, visit <a href="https://skyux.developer.blackbaud.com/">skyux.developer.blackbaud.com/</a>.
```
Blackbaud-SteveBrush: Could it be simplified a bit? Something like, "This site describes the version of SKY UX that supports the latest version of Angular. For information about the version of SKY UX that supports the AngularJS framework, visit..."

blackbaud-johnly: Sounds good to me. I wasn't sure if "the latest version of Angular" was correct, which is why I had the longer, more unwieldy wording in place before. But if this shorter version is accurate, I'm all for it. I've updated the text with our suggested version. Thanks!

PR #186, opened 2018-12-12
[HOLD] progress indicator examples

Blackbaud-ToddRoberts:


Blackbaud-MattGregg:
@Blackbaud-ToddRoberts I think the Anatomy section would be more useful ahead of the behavior section. The behavior section seems to reference or assume someone already knows the anatomical parts. 

This sentence in the first usage section seems incomplete or fragment "Actions that can be taken in bulk on selected items in a multiselect." Wondering if this can be clearer. Something like Actions can be taken on multiple items in a list of items. 

Also wondering why the split panel is listed as related but isn't linked and included in the usage section given it doesn't really exist in SKY UX yet and we haven't moved to new content strategy? 

Can this be used in the Flyout if someone is using a list in there? Thinking yes, but could be good to call this out. Related, not advised to use in tiles?

Blackbaud-ToddRoberts:
@Blackbaud-MattGregg I assume this feedback is intended for #182 ?

Blackbaud-MattGregg:
@Blackbaud-ToddRoberts Yep. Moved it over there. Too many tabs I guess ;)

Blackbaud-ToddRoberts:
closing in favor of #625

PR #188, opened 2018-12-13
Added beforeClose event to modal docs

blackbaud-conorwright:
Issue: blackbaud/skyux2#2138

### DIFF
--- a/src/app/components/modal/index.html
+++ b/src/app/components/modal/index.html
@@ -120,7 +120,7 @@
     <sky-demo-page-property
       propertyName="beforeClose"
     >
-      An event that the modal instance emits when it is about to close. It emits a <stache-code>SkyModalBeforeCloseHandler</stache-code> object with a <stache-code>closeModal</stache-code> method that completes the close of the modal. The modal will not close unless the <stache-code>closeModal</stache-code> method is called if there is a subscription to this event.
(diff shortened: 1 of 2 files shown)


src/app/components/modal/index.html
```diff
@@ -118,6 +117,11 @@
   </sky-demo-page-properties>
 
   <sky-demo-page-properties sectionHeading="Modal instance events">
+    <sky-demo-page-property
+      propertyName="beforeClose"
+    >
+      An event that the modal instance emits when it is about to close. It emits a <stache-code>SkyModalBeforeCloseHandler</stache-code> object with a <stache-code>closeModal</stache-code> method that completes the close of the modal. The modal will not close unless the <stache-code>closeModal</stache-code> method is called if there is a subscription to this event.
```
blackbaud-johnly: In the second sentence, would it change the meaning if we change "that completes the close of the modal" to "that closes the modal"?

In the last sentence, I'd suggest moving the if clause to the start of the sentence and tweaking to use active voice: "If a subscription exists for this event, the modal does not close until that description calls the `closeModal` method." Does that work?

blackbaud-conorwright: *subscriber calls the

But yes these changes look good 🙂

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/2138.

PR #211, opened 2019-01-17
Document the ListItemModel object on list filters

Blackbaud-TrevorBurch:


src/app/components/list/filters/index.html
```diff
@@ -69,48 +67,66 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="id"
+    >
+    Specifies the ID for the item in the list. This property accepts <stache-code>string</stache-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="isSelected"
+      isOptional="true"
+      defaultValue="false"
```
blackbaud-johnly: Is this default correct? Based on https://github.com/blackbaud/skyux-list-builder-common/blob/master/src/app/public/state/items/item.model.spec.ts#L22, I assume the default should be `true`.

src/app/components/list/filters/index.html
```diff
@@ -69,48 +67,66 @@
...
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListItemModel properties">
+    <sky-demo-page-property
+      propertyName="data"
+    >
+    Specifies the data for the item in the list.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="id"
```
blackbaud-johnly: Should we indicate that this property is required? Based on https://github.com/blackbaud/skyux-list-builder-common/blob/master/src/app/public/state/items/item.model.ts#L8, I think it might be required.

src/app/components/list/filters/index.html
```diff
@@ -69,48 +67,66 @@
...
-    Specifies the text for the filter's summary item. If not specified, then the filter summary item text defaults to the <stache-code>value</stache-code> of the filter.
+    Specifies the value of the filter.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListItemModel properties">
+    <sky-demo-page-property
+      propertyName="data"
+    >
+    Specifies the data for the item in the list.
```
blackbaud-johnly: Can we be any more specific than "data"? Or does this property include any and all data within list items?

src/app/components/list/filters/index.html
Blackbaud-TrevorBurch: No. That link is setting the value to `true` on line 20 but the default is `false`.

src/app/components/list/filters/index.html
Blackbaud-TrevorBurch: Yes. Good catch.

src/app/components/list/filters/index.html
Blackbaud-TrevorBurch: It is any and all data. So it is whatever data has came in from consumers. We don't really have any more knowledge than that.

PR #217, opened 2019-01-24
Document pipelineSettings property

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/2276 for initial documentation on dev work in https://github.com/blackbaud/skyux-config/pull/13/files. (The dev work will require further documentation for https://github.com/blackbaud/skyux2/issues/2277 down the road.)

### DIFF
--- a/src/app/learn/reference/configuration/index.html
+++ b/src/app/learn/reference/configuration/index.html
@@ -260,6 +260,11 @@
     </li>
     <li>
       <p>
+        <stache-code>pipelineSettings</stache-code> — Specifies configuration options for build and release pipelines.
+      </p>
+    </li>
+    <li>
+      <p>
         <stache-code>plugins</stache-code> — Specifies plugins to expand on existing SKY UX Builder functionality and create new functionality. For example, the Stache 2 component library uses <a href="https://developer.blackbaud.com/stache/learn/plugins">a variety of plugins</a>.
       </p>
     </li>


src/app/learn/reference/configuration/index.html
```diff
@@ -258,6 +258,11 @@
         </li>
       </ul>
     </li>
+    <li>
+      <p>
+        <stache-code>pipelineSettings</stache-code> — Specifies configuration options for build and release pipelines. The data type of this property is <stache-code>any</stache-code>. This property is for internal Blackbaud use only.
```
blackbaud-johnly: @Blackbaud-BobbyEarl, I added a couple sentences after talking with @Blackbaud-SteveBrush about this property. Please let me know what you think. Thanks.

Blackbaud-BobbyEarl: What do you think about the following?

```
Specifies configuration options for build and release pipelines. The data type of this property is <stache-code>any</stache-code>. Users are free to use this property as they see fit to control their build process.  Internally, Blackbaud is using it in such a manner.
```

Even writing that I'm not a huge fan.  I'm trying to convey that it's not necessarily a Blackbaud only property, meaning anyone can use it for whatever they want (hence the `any` type), but that Blackbaud is using it internally.

blackbaud-johnly: What do you think of replacing my last sentence and your last 2 sentences with: "SKY UX does not provide public access to a pipeline, but Blackbaud uses this property internally to control our build processes. External consumers can use this property to define their own pipeline values for their build machines, but SKY UX not support those build processes beyond providing this container for custom values."

PR #224, opened 2019-01-30
Confirm service doc updates

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2/issues/1847 and also https://github.com/blackbaud/skyux2/issues/2289.


src/app/components/confirm/index.html
```diff
@@ -5,95 +5,76 @@
...
+    The confirm service launches simple confirmation dialogs to allow users to confirm actions. The <stache-code>SkyConfirmService</stache-code> provider launches the confirmation dialog. Within the service, you can specify the dialog's message and customize the button text.
   </sky-demo-page-summary>
 
-  <sky-demo-page-properties sectionHeading="Service methods">
+  <sky-demo-page-properties sectionHeading="Confirm service methods">
     <sky-demo-page-property
       propertyName="open(config)"
     >
-    Opens a confirmation dialog. The <stache-code>config</stache-code> parameter is of type <stache-code>SkyConfirmConfig</stache-code> and passed to the component's constructor. This method returns the confirmation dialog instance.
+    Opens a confirmation dialog and passes the <stache-code>config</stache-code> parameter is of type <stache-code>SkyConfirmConfig</stache-code> to the component's constructor. This method returns the confirmation dialog instance.
```
Blackbaud-SteveBrush: "...and passes the config parameter is of type..." Sounds strange?

Blackbaud-SteveBrush: Also, would it help to include the specific returned type? "This method returns an object of type `SkyConfirmInstance`." Then, we can document the properties and methods of the `SkyConfirmInstance` in the section below, that is currently titled "Confirmation dialog events".

src/app/components/confirm/index.html
```diff
@@ -5,95 +5,76 @@
...
+      </li>
+      <li>
+        <stache-code>YesNoCancel</stache-code> — Displays three buttons with <strong>Yes</strong>, <strong>No</strong>, and <strong>Cancel</strong> as the default labels. The integer value equivalent is 3.
+      </li>
+    </ul>
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
-  <sky-demo-page-properties sectionHeading="Events">
+  <sky-demo-page-properties sectionHeading="Confirmation dialog events">
```
Blackbaud-SteveBrush: It might be beneficial to document the type of `SkyConfirmInstance` here, instead of isolating this section to "events", since we might add properties and methods to the instance in the future. I'm unsure of your current docs patterns, but something like the following?
```
<sky-demo-page-properties sectionHeading="Confirmation dialog instance (SkyConfirmInstance)">
```

src/app/components/confirm/index.html
blackbaud-johnly: For the first point, I think the problem is just the stray "is" that needed to be deleted. So I deleted that. ... For the second point, I replaced the existing second sentence with your suggestion and changed the "Confirmation dialog events" header to "SkyConfirmInstance events"

src/app/components/confirm/index.html
blackbaud-johnly: I changed this to "SkyConfirmInstance events"

blackbaud-johnly: I also fleshed out the description of the `closed` property to reference the `SkyConfirmCloseEventArgs` object and document its `action` property.

PR #228, opened 2019-02-05
Document blur event on select field

blackbaud-conorwright:
relates to issue: blackbaud/skyux-select-field#10

I also added doc changes for https://github.com/blackbaud/skyux2/issues/2297


src/app/components/select-field/index.html
```diff
@@ -6,86 +6,93 @@
...
-    Specifies the ghost text to display in the select field appears when the <stache-code>selectMode</stache-code> is <stache-code>single</stache-code> and no item has been selected.
+    Specifies ghost text to display in the text input when <stache-code>selectMode</stache-code> is set to <stache-code>single</stache-code> and no item is selected. This property accepts <stache-code>string</stache-code> values.
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
+  <sky-demo-page-properties sectionHeading="Select field events">
+    <sky-demo-page-property
+      propertyName="blur"
+    >
+      Fires when the select field component's text input or button is touched. This event does not emit a value.
```
blackbaud-johnly: Can we change "touched" to "selected"? Or would that change the meaning?

blackbaud-johnly: @Blackbaud-TrevorBurch and @blackbaud-conorwright, now thay https://github.com/blackbaud/skyux-select-field/pull/11/files is merged, can we update the docs? Or is this still pending?

blackbaud-conorwright: Touched/blurred have weird meaning. How about `when the select field component loses focus`? since that is the actual trigger now

blackbaud-johnly: Works for me. Anything else need to be documented after all the changes?

blackbaud-conorwright: I don't believe so. Should be good to go once the skyux2 release is out

PR #231, opened 2019-02-08
Inline form docs

Blackbaud-ToddRoberts:


Blackbaud-AlexKingman:
@Blackbaud-ToddRoberts This is great. FYI - I have a PR incoming that contains all the property definitions and a demo.

src/app/components/demo-components.service.ts
```diff
@@ -198,6 +198,13 @@ export class SkyDemoComponentsService {
         summary: 'The infinite scroll component dynamically loads data as users scroll.',
         url: '/components/infinite-scroll',
         getCodeFiles: () => this.getDemoFiles('SkyInfiniteScrollDemoComponent')
+      },
+			{
+        name: 'Inline form',
+        icon: 'keyboard-o',
+        summary: 'The inline form component renders a form in context rather than in a modal.',
```
blackbaud-johnly: In the corresponding file, we say "in the current view" rather than "in context." I think we should be consistent. Personally, I think "in the current view is clearer. But either way, I'd use the same wording in both places.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+	
+<sky-demo-page-content sectionHeading="UX guidelines">
+
+<header><div class="sky-section-heading">Usage</div></header>
+	<header><div class="sky-subsection-heading sky-margin-stacked-default">When to use inline forms</div></header>
+	<stache-row>
+    <stache-column screenSmall="6">
+			<ul>
+			<li>The add/edit form is simple in terms of a small # of fields and low complexity of input controls</li>
+			<li>The surrounding content will not distract the user from the add/edit task</li>
```
blackbaud-johnly: I'd change "the user" to "users." Also, I 'd change "add/edit" to "add or edit."

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>There is a high likelihood the user would need to add multiple records in succession</li>
```
blackbaud-johnly: Not a huge deal, but I avoid "There is" and "There are" sentence constructions whenever possible. It usually leads to passive voice. (Although sometimes it can't be avoided.) ... I'd suggest changing this to "Users are highly likely to add multiple records in succession."

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>Adding/editing a subsection within a larger modal form</li>
```
blackbaud-johnly: I'd change "Adding/editing" to "To add or edit."

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>It’s ok if the user abandons the task while the add/edit form is open (i.e. the task is functionally non-modal)</li>
```
blackbaud-johnly: Might just be me, but I'm not a huge fan of the "It's ok if" construction. I'd flip the order around a bit to use a more active voice: "Users can abandon the task... I'd also change "add/edit" to "add or edit." ... I'm also not a huge fan of the parenthetical at the end. Can we change it to "...is open because the task is functionally non-modal."?

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+		</stache-column>
+	</stache-row>
+		
+	
+	
+	<header><div class="sky-subsection-heading sky-margin-stacked-default">When not to use inline forms</div></header>
+	<stache-row>
+    <stache-column screenSmall="6">
+			<ul>
+			<li>Add/edit form would lead to a large amount of page morphing or a user needing to scroll to complete the form if done inline</li>
```
blackbaud-johnly: I'd change "add/edit" to "The add or edit." ... I'd change "would lead to" to "requires" and change "a user needing" to "users must." ... Also, I don't think we need "if done inline" at the end. This seems redundant, and I'd delete it.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>Working with a paginated or infinite scroll list of items</li>
```
blackbaud-johnly: I'd change "Working with" to "Within" ... Also, should we link to the paging, infinite scroll, or list components? I assume not, but just making sure.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>You want the user to have to finish the add/edit before doing anything else (i.e. the task is functionally modal)</li>
```
blackbaud-johnly: I'd change "You want the user to have to" to "Users must" (unless we don't want to state this rule too strongly). ... I'd change "the add/edit" to "the add or edit form." ... I'm also not a huge fan of the parenthetical at the end. Can we change it to "...anything else because the task is functionally modal."?

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<div class="screenshot screenshot-info">
+				<img src="~/assets/img/guidelines/inline-form/anatomy.png" style="width: 100%;" />
+			</div>
+		</stache-column>
+	</stache-row>
+		
+	
+	<header><div class="sky-section-heading">Behavior</div></header>
+		<header><div class="sky-subsection-heading sky-margin-stacked-default">Add form</div></header>
+			<p>When the user clicks the Add button, the form should slide out from the bottom of the list.</p>
```
blackbaud-johnly: I'd change "the user" to "users." ... I think "should slide" should be "slides" to make it a rule, not a suggestion.

src/app/components/demo-components.service.ts
Blackbaud-ToddRoberts: "current view" wording seems better to me as well.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+	<p>The inline form component renders a form in the current view rather than opening a modal.</p>
+</sky-demo-page-summary>
+	
+<sky-demo-page-content sectionHeading="UX guidelines">
+
+<header><div class="sky-section-heading">Usage</div></header>
+	<header><div class="sky-subsection-heading sky-margin-stacked-default">When to use inline forms</div></header>
+	<stache-row>
+    <stache-column screenSmall="6">
+			<ul>
```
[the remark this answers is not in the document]
Blackbaud-ToddRoberts: ok, makes sense.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,89 @@
...
+			<li>The add/edit form is simple in terms of a small # of fields and low complexity of input controls</li>
```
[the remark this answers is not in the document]
Blackbaud-ToddRoberts: new wording is good with me.

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: ok

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: ok makes sense

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: ok

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: Can probably just remove the parenthetical part. Reordering the wording is fine, intent is to communicate that it won't cause problems if the user abandons the task.

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: these all make sense

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: don't think we need to link to paging/infinite scroll

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: Can probably remove the parenthetical part as in earlier comment. other changes seem good.

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: ok

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,155 @@
...
+	<sky-demo-page-property
+    propertyName="buttonLayout"
+  >
+  Specifies the buttons to display with the inline form. This property accepts a <stache-code>SkyInlineFormButtonLayout</stache-code> value. The valid options are <stache-code>Custom</stache-code> for custom buttons, <stache-code>DoneCancel</stache-code> for Done and Cancel buttons, <stache-code>DoneDeleteCancel</stache-code> for Done, Delete, and Cancel buttons, <stache-code>SaveCancel</stache-code> for Save and Cancel buttons, and <stache-code>SaveDeleteCancel</stache-code> for Save, Delete, and Cancel buttons.
+  </sky-demo-page-property>
+	<sky-demo-page-property
+    propertyName="buttons"
+    isOptional="true"
+  >
+  Specifies configuration options for the inline form's buttons. This property accepts <a stacheRouterLink="." fragment="skyinlineformbuttonconfig-properties">a <stache-code>SkyInlineFormButtonConfig</stache-code> object</a>.
```
Blackbaud-AlexKingman: This property accepts an *array* of `SkyInlineFormButtonConfig` objects.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,155 @@
...
+<sky-demo-page-properties sectionHeading='Inline form properties'>
+	<sky-demo-page-property
+    propertyName="config"
+  >
+  Specifies configuration options for the buttons to display with the inline form. This property accepts <a stacheRouterLink="." fragment="skyinlineformconfig-properties">a <stache-code>SkyInlineFormConfig</stache-code> object</a>.
+  </sky-demo-page-property>
+	<sky-demo-page-property
+    propertyName="showForm"
+  >
+  Specifies the form to display inline.
```
Blackbaud-AlexKingman: Specifies when the inline form should display (or something like that). In other words, this property toggles the display of the form.
Type: boolean.

src/app/components/inline-form/index.html
```diff
@@ -0,0 +1,155 @@
...
+  >
+  Specifies a template to use to instantiate the inline form. This property accepts <a href="https://angular.io/api/core/TemplateRef">an Angular <stache-code>TemplateRef</stache-code> class</a>.
+  </sky-demo-page-property>
+</sky-demo-page-properties>
+
+<sky-demo-page-properties sectionHeading='Inline form events'>
+	<sky-demo-page-property
+    propertyName="close"
+  >
+  Fires when users close the inline form.
```
Blackbaud-AlexKingman: Should this also say it emits a type of `SkyInlineFormCloseArgs`?

PR #232, opened 2019-02-08
added note about required and excludeFromRequest

Blackbaud-DeniseCisneros:
Bobby, will you confirm if we want to include `$schema` after all and also check if that's what we want to say about `required` and `excludeFromRequest`?


Blackbaud-BobbyEarl:
## Regarding `required`

I couldn't clearly remember, so I had to dig in a bit around the `required` parameter.  Consumers have the ability to mark any parameter as `required`.  

### For `leid`, `envid`, and `svcid`

For these parameters, `@blackbaud/skyux-builder` and `@skyux-sdk/builder` passes the required state to `auth-client`, when then in turn attempts to resolve the default value (if only one is available), show the welcome screen (if multiple are available), or show an error (if none are available).

- https://github.com/blackbaud/skyux-builder/blob/master/runtime/bootstrapper.ts
- https://github.com/blackbaud/skyux-sdk-builder/blob/master/runtime/bootstrapper.ts
- https://github.com/blackbaud/auth-client/blob/master/src/context/context-provider.ts

### For any custom required params

No logic is applied; however, consumers would be free to use the `SkyuxConfigParams` class to interact with them.

- https://github.com/blackbaud/skyux-config/blob/eabecfc6630dd01cdcd3efe68849b3f0b20562e1/src/app/public/params.ts#L113-L134

## Regarding `excludeFromRequest`

I think we can elaborate slightly on what this means.  By default, any `param` listed which contains a value will be included in requests made from `SkyAuthHttp` and `SkyAuthHttpClient`.  If, for some reason, a consumer wants to track a param but not have it automatically included in those http calls, this parameter is applicable.

- https://github.com/blackbaud/skyux-http/blob/master/src/app/public/modules/auth-http/auth-http.ts#L101-L114
- https://github.com/blackbaud/skyux-http/blob/master/src/app/public/modules/auth-http/auth-http.ts#L101-L114
- https://github.com/blackbaud/skyux-config/blob/master/src/app/public/params.ts#L190-L205

blackbaud-johnly:
Also addresses https://github.com/blackbaud/skyux2/issues/1838.

PR #233, opened 2019-02-11
Add page anchors to configuration reference

blackbaud-johnly:


Blackbaud-BobbyEarl:
@blackbaud-johnly I really like this!  Thank you for taking on the task.

One quick comment.  I like how this makes the main sections linkable.  Those sections are much smaller, but the child properties have the same problem this is solving.  Do you have any thoughts on flattening the structure, so something like `app`'s children would become top-level, but listed as `app.externals`, `app.port`, `app.title`, etc.?

I've gone back and forth on my thoughts about it.

blackbaud-johnly:
@Blackbaud-BobbyEarl, I went back and forth on the child properties too. I decided not to make them page anchors to keep the list a little smaller and to make it clear at a glance that those properties are children of their parents. I figured we could try it this way first to gauge the effectiveness and then flatten the structure in a future update if we decide it's warranted. ... That said, I don't feel strongly on this. If you want, I can go ahead and flatten the structure like you suggest. Just let me know.

Issue #310, opened 2018-01-10
Add page title guideline to Tabs component page


Blackbaud-ToddRoberts:
When using tabs as main navigation, the tab titles should be reflected in the page title.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/323.

PR #323, opened 2019-02-27
Add page title guideline to tab docs

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/310.


blackbaud-johnly:
I added that section @Blackbaud-ToddRoberts. What do you think?

Issue #303, opened 2018-05-15
Update UX guidelines for modal component


blackbaud-johnly:
Got feedback from Ben L that he'd like the UX Guidelines section for the modal component to include prescriptive notes like:
o	use sky-btn-link for Cancel buttons
o	if only 1 button, use "Close" and primary (I think?)
o	header captions should use (<decide on sentence/proper/etc>) case (ex: "Add new constituent" or "Add New Constituent")

I brought this up with Todd, and he said it is worth adding these things. Adding that he also needs to update the examples in the UX guidelines for modals. Those examples should have a primary Save button and a Cancel link. In general, if a modal only has one action it should use a Confirm instead.

PR #325, opened 2019-02-27
added guidelines for buttons and titles

Blackbaud-DeniseCisneros:
In reference to issue https://github.com/blackbaud/skyux2-docs/issues/303


src/app/components/modal/index.html
```diff
@@ -150,26 +150,15 @@
...
-			</div>
-  	<div class="sky-subsection-heading">Usage</div>
-			<div class="sky-emphasized">Modal</div>
+		<div class="sky-section-heading">Usage</div>
+			<div class="sky-subsection-heading">Modal</div>
 			<p>A modal dialog is appropriate for forms that meet any of the following criteria:</p>
 			<ul>
 				<li>The task is functionally modal, and users must finish it before doing anything else.</li>
 				<li>Displaying the form inline would lead to substantial content shifting or require users to scroll to complete the form.</li>
 				<li>The form adds records from a paginated list or an infinite scroll list.</li>
```
blackbaud-johnly: This third bullet is not reflected in the corresponding guidelines at https://developer.blackbaud.com/skyux/design/guidelines/form-design#form-containers. It seems like they should be the same. @Blackbaud-ToddRoberts, is there any reason to exclude this bullet from that corresponding list? If not, should we go ahead and add it as part of this PR?

blackbaud-johnly: Also, should we link "paginated list" to the paging component and link "infinite scroll list" to the infinite scroll component?

src/app/components/modal/index.html
```diff
@@ -179,6 +168,25 @@
 				<li>The form supports a complex task for a list of objects such as filling out report cards for a class.</li>
```
blackbaud-johnly: Similar question here. The fourth bullet does not match the corresponding fourth bullet at https://developer.blackbaud.com/skyux/design/guidelines/form-design#form-containers. All the other 5 bullets match up. Should this bullet match as well? If so, how? @Blackbaud-ToddRoberts, should this bullet be copied to that corresponding list and the fourth bullet there copied here? Of should one of the fourth bullets replace the other? Seems like something is out of whack here, but I'm not sure how to fix it.

src/app/components/modal/index.html
```diff
@@ -179,6 +168,25 @@
...
+			</div>
+			<p>If there is not an indirect object or the user is the implied indirect object, you can exclude it:</p>
+			<p><span class="sky-emphasized">&lt;Verb&gt; &lt;direct object&gt;</span></p>
+			<div style="margin-bottom: 20px;">
+				<img class="screenshot screenshot-info" src="~/assets/img/guidelines/wording/modal-without-object.png" alt="modal title: Manage bank feeds"/>
+			</div>
+		
+		<div class="sky-subsection-heading">Buttons in the modal action bar</div>
+			<ul>
+				<li>Use a primary <a routerLink="/components/button">button</a> for the most likely action the user will take. Use secondary buttons for less likely actions. Use a link button for the Cancel action.</li>
```
blackbaud-johnly: Might just be me, but I think "most likely action" reads awkwardly. I'd change this to "... for the most important action. Use secondary buttons for less important actions..."

src/app/components/modal/index.html
Blackbaud-ToddRoberts: I don't think we need to link to paging/infinite scroll here, it's such ancillary info.

Can add the third bullet from modal to form design, was probably a copypasta error.

src/app/components/modal/index.html
Blackbaud-ToddRoberts: that sounds good to me.

src/app/components/modal/index.html
Blackbaud-ToddRoberts: I'm not sure which wording on those fourth bullets is better - they are trying to convey the same info. I assume one of them was wordsmithed, so whichever one you prefer can be the winner.

src/app/components/modal/index.html
```diff
@@ -179,6 +168,25 @@
...
+				<li>If a modal has only one action, you should generally use the <a routerLink="/components/confirm">Confirm</a> component instead.</li>
```
blackbaud-johnly: "confirm" should be lowercase. (Also, no big deal, but I'd include "the" and "component" as part of the link text so that it stands out a bit more. But that might just be me.) ... 

Also, can we delete "you should generally" to word this more strongly? Maybe there are exceptions to this rule, but if it's worth calling out, I think we should state it as a hard and fast rule rather than a recommendation. Again, just my 2 cents.

src/app/components/modal/index.html
blackbaud-johnly: Great. I'll add that third bullet to the other guidelines and push it up here.

src/app/components/modal/index.html
Blackbaud-ToddRoberts: There are exceptions for cases where there needs to be a lot of content on the modal.

Is something like "try using the Confirm component" still too wishy washy?

blackbaud-johnly: I like that better, but if we want to avoid stating this as a hard and fast rule, I think either is fine. i made the (incorrect) assumption that the wording was an oversight and not an intentional thing. Sorry about that.

src/app/components/modal/index.html
blackbaud-johnly: @Blackbaud-DeniseCisneros, can you update this to consistently use whichever text you prefer? (Just my 2 cents, but I prefer the wording at https://developer.blackbaud.com/skyux/design/guidelines/form-design. But it's up to you which you think we should use.)

Blackbaud-DeniseCisneros: I like the wording on the form design page too and updated the bullet with that text.

src/app/components/modal/index.html
Blackbaud-DeniseCisneros: ✔️

src/app/components/modal/index.html
Blackbaud-DeniseCisneros: I updated casing but left the verbiage as is. Let me know if that should've been changed after all.

src/app/components/modal/index.html
blackbaud-johnly: Looks good. Thanks!

src/app/components/modal/index.html
blackbaud-johnly: LGTM

PR #331, opened 2019-03-01
added alert with link to bb help docs site

Blackbaud-DeniseCisneros:
Added a link to the Help Widget docs

References issue https://github.com/blackbaud/skyux2/issues/2292


src/app/components/help-inline/index.html
```diff
@@ -8,6 +8,12 @@
     <p>
       The help inline component creates a small button next to a field. When clicked they can trigger an action that displays more information about that field that the user might need information on.
     </p>
+
+    <p>
+      <sky-alert alertType="info">
+        <i class="fa fa-lock" aria-hidden="true"></i> <b>Blackbaud Staff:</b> To implement the Help Widget in a core application or SPA, see <a stacheRouterLink="https://docs.blackbaud.com/bb-help-docs/">BB Help Services</a>.
```
blackbaud-johnly: I think we should flesh this out a bit and tweak the wording slightly to be consistent with the other places on the SKY UX site where we reference Blackbaud-only features: "To display context-sensitive help content in a flyout panel, Blackbaud developers can use <a stacheRouterLink="https://docs.blackbaud.com/bb-help-docs/">the Help Widget</a>. This feature is for internal Blackbaud use only."

Also, what do you think of making this a regular paragraph instead of a note? Based on the design team's reluctance about how we use alerts, I try to avoid them except in very rare cases. I don't think this content really needs to be highlighted this strongly. If you disagree, keep it as a note. But if not, I think it works just as well as an add-on to the existing overview paragraph.

blackbaud-johnly: The edit doesn't reflect it, but I was suggesting making "the Help Widget" the text to use as a link to that site.

Blackbaud-DeniseCisneros: Thanks for that, John! I made those updates.

PR #354, opened 2019-03-15
Flyout, summary action bar, grid guideline updates

Blackbaud-ToddRoberts:


src/app/components/flyout/flyout-guidelines.component.html
```diff
@@ -127,7 +127,7 @@
 	
 	<stache-row>
     <stache-column screenSmall="6">
-			<p>Do not use a flyout for tasks where users always need a full-screen view of details, such as expanded views of tiles. Navigate to a full page instead.</p>
+			<p>Do not use a flyout for tasks where users will need a full-screen view of details or will be taking multiple actions within the view. Navigate to a full page instead.</p>
```
Blackbaud-DeniseCisneros: Because we're trying to move away from using "do not," maybe something like:

Use a full a page when users need a full-screen view of details or when users can perform multiple actions within the view.

src/app/components/flyout/flyout-guidelines.component.html
```diff
@@ -15,7 +15,7 @@
 	
 	<stache-row>
     <stache-column screenSmall="6">
-			<p>Use a flyout to open records in a list so that users can return to the list view after completing quick tasks on the record pages.</p>
+			<p>Use a flyout to open records in a list so that users can return to the list view after completing quick tasks on the record pages. If the user will need a full-screen view of details or will be taking multiple actions on the record, navigate to a full page instead.</p>
```
Blackbaud-DeniseCisneros: The new sentence also appears in the When not to use section. I'm not sure if we should keep it to really drive the point home, or if it just comes across as redundant.

Blackbaud-ToddRoberts: It started in the Do Not section but got overlooked so we wanted to double up to ensure it gets across

blackbaud-johnly: I'd tweak the opening clause to make "user" plural and use present tense verbs: "If users require a full-screen view of details or need to take multiple actions on the record,"

src/app/components/flyout/flyout-guidelines.component.html
```diff
@@ -105,7 +105,7 @@
 	
 	<stache-row>
     <stache-column screenSmall="6">
-			<p>Do not use a flyout to display items that need a long time to load such as system-generated reports. Instead, let users continue to use the system, and use <a routerLink="../alert">an alert component</a> to let them know when the content is available.</p>
+			<p>Do not use a flyout to display items that need a long time to load such as system-generated reports. Instead, let users continue to use the system, and use <a routerLink="../toast">an toast</a> to let them know when the content is available.</p>
```
blackbaud-johnly: "an" before "toast" should be "a"

src/app/components/flyout/flyout-guidelines.component.html
blackbaud-johnly: I'd make similar changes here to use present tense verbs: "Do not use a flyout for tasks where users require a full-screen view of details or need to take multiple actions in the view..."

src/app/components/grid/index.html
```diff
@@ -257,6 +257,9 @@
 					<li>
 						<span class="sky-deemphasized">Toolbar (optional)</span>
 					</li>
+					<li>
+						<span class="sky-deemphasized">Multiselect toolbar (optional)</span>
```
blackbaud-johnly: When I serve locally, I don't see either of these changes to the grid topic. Do you see them in place when you serve locally? I saw the changes to the flyout topic, so I don't know why I wouldn't see these.

blackbaud-johnly: Nevermind. I figured out what the problem is....

Blackbaud-ToddRoberts:
I just pushed some edits based on feedback

src/app/components/summary-actionbar/index.html
```diff
@@ -84,6 +84,7 @@
   <stache-row>
     <stache-column screenSmall="6">
 			<p>Actions that can be taken on multiple items in a list of items.</p>
+			<p>When used with a multiselect list, hide the summary action bar until the user selects an item in the list. Once an item is selected the summary action bar should fly in from the bottom of the page.</p>
```
blackbaud-johnly: I'd change "the user selects" to "users select." And I'd change "Once" to "After" (it's a translation/localization thing). Also, we should have a comma after the opening clause in the second sentence, but I'd also reword the opening clause to use active voice: "After users select an item, the summary..." And should "should fly in" be "flies in"? I mean, is there an alternative to flying in from the bottom or the page?

src/app/components/grid/index.html
```diff
@@ -321,6 +324,7 @@
 			<ul>
 				<li>The list is presented in a tile, page section, or other container that might flow off the screen.</li>
 			</ul>
+			<p>Include the multiselect toolbar to allow users to select or clear all items and optionally view only the selected items.</p>
```
blackbaud-johnly: I'd delete "optionally" and replace it with "to" since that's implicit from the fact that toolbar allows users to do it that this is optional. ... Also, it might just be me, but wording this sentence as a command seems odd here. It seems like it should just be a sentence about what the toolbar does: "The multiselect toolbar allows users to..."

src/app/components/grid/index.html
```diff
@@ -321,6 +324,7 @@
...
       <p>When users select rows, the summary action bar should fly in from the bottom of the page. Place actions that can be taken on the selected items in the summary action bar.</p>
```
blackbaud-johnly: I know this isn't new content, but I have the same question here as in the comment on the summary action bar change. Should "should fly in" here be "flies in"?

src/app/components/summary-actionbar/index.html
Blackbaud-ToddRoberts: Right now the flyin is not built into the component. Not sure what our standards are around wording things that are part of a pattern but need to be manually implemented.

blackbaud-johnly: OK. Thanks.

src/app/components/grid/index.html
blackbaud-johnly: Resolving based on response to the other comment.

PR #377, opened 2019-03-28
Redo tab & popover pages

Blackbaud-ToddRoberts:


src/app/components/popover/index.html
```diff
@@ -1,13 +1,161 @@
...
+		
+		<sky-demo-page-content sectionHeading="Options">
+			<h3>Positioning</h3>
+			<div class="two-col-right-image sky-margin-stacked-separate">
+				<div class="two-col-text">
+					<p>You can specify the position of the popover in relation to the element it is pointing to. The default position is above and centered on the element.</p>
+				</div>
+				<div class="two-col-image">
+					<div class="screenshot">
+						<img src="~/assets/img/guidelines/popover/option-position.png" />
```
blackbaud-johnly: Should the image have a caption? I don't think so because it **seems** like the pattern we are following is to provide captions only when the images illustrate something to do or not to do. But I can't tell for certain. Is my assumption about the pattern correct?

src/app/components/popover/index.html
```diff
@@ -1,13 +1,161 @@
...
+		
+		<sky-demo-page-content sectionHeading="Layout">
+			<div class="two-col-right-image sky-margin-stacked-separate">
+				<div class="two-col-text">
+					<p>In most cases, use the default “above” placement for Popovers. This minimizes the chances that the Popover will obscure content that a user hasn’t seen yet or will interact with next.</p>
+					<p>Use other placement options if the trigger for the Popover will be close to the edge of the display. Choose a placement that will obscure as little other content as possible.</p>
+				</div>
+				<div class="two-col-image">
+					<div class="screenshot sky-margin-stacked-content-section">
+						<img src="~/assets/img/guidelines/popover/layout-alternate.png" />
```
blackbaud-johnly: Should this image have a caption? If I'm interpreting the caption correctly, it seems like this should have a caption along the lines of "Do use a placement option that minimizes the risk of obscuring content." Or something like that.

src/app/components/shared/demo-page-properties.component.html
```diff
@@ -1,6 +1,12 @@
 <sky-demo-page-content [sectionHeading]="sectionHeading">
   <ng-content></ng-content>
-  <ul>
-    <ng-content select="sky-demo-page-property"></ng-content>
-  </ul>
-</sky-demo-page-content>
+  <div class="property-table" role="table">
+			<div class="property-table-row property-table-header" role="row">
+				<div class="property-name" role="columnheader">Property name</div>
```
blackbaud-johnly: Does this header work for all scenarios? For example, does "property name" work for the tables with lists of events? i think it's probably fine, but I just wanted to raise the question to make sure.

src/app/components/tabs/index.html
```diff
@@ -1,15 +1,292 @@
...
+	<sky-demo-page-content sectionHeading="Anatomy">
+		<div class="two-col-right-image sky-margin-stacked-separate">
+			<div class="two-col-text">
+				<div class="anatomy-list">
+					<div class="anatomy-item anatomy-item-required"><div class="anatomy-marker">1</div><div class="anatomy-description">Selected tab</div></div>
+					<div class="anatomy-item anatomy-item-required"><div class="anatomy-marker">2</div><div class="anatomy-description">Unselected tab</div></div>
+					<div class="anatomy-item anatomy-item-required"><div class="anatomy-marker">3</div><div class="anatomy-description">Tab border</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">4</div><div class="anatomy-description">Tab count (optional)</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">5</div><div class="anatomy-description">Close tab button (optional)</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">6</div><div class="anatomy-description">Open tab button (optional)</div></div>
```
blackbaud-johnly: I think we are missing an item here. We have the Open and Close buttons, but not the Add button. That's a separate button, right? I see it in the image below for the Open and close section and I see a property for it in the properties for the components, so I believe so.

src/app/components/tabs/index.html
```diff
@@ -1,15 +1,292 @@
...
+	
+	<sky-demo-page-content sectionHeading="Options">
+		<h3>Count</h3>
+			<div class="two-col-right-image sky-margin-stacked-separate">
+				<div class="two-col-text">
+					<p>When you use tabs to hold lists, you can include the list's record count in the tab so users know if there are records to look at without navigating to the tab.</p>
+				</div>
+				<div class="two-col-image">
+					<div class="screenshot">
+						<img class="bordered" src="~/assets/img/guidelines/tab/option-count.png" />
```
blackbaud-johnly: Should this image have a caption? It seems like maybe it should have something along the lines of "Do display record counts in tabs that contain lists."

src/app/components/tabs/index.html
```diff
@@ -1,15 +1,292 @@
...
+				</div>
+			</div>
+		<h3>Open and close</h3>
+			<div class="two-col-right-image sky-margin-stacked-separate">
+				<div class="two-col-text">
+					<p>If users need to work with a subset of large number of available tabs, let them open and close tabs so they are only working with relevant tabs. If users have more than one option of what to display in a new tab, use action buttons within the tab view to let them choose the option they want.</p>
+				</div>
+				<div class="two-col-image">
+					<div class="screenshot">
+						<img class="bordered" src="~/assets/img/guidelines/tab/option-open-close.png" />
```
blackbaud-johnly: Should this image have a caption? It seems like maybe it should have something along the lines of "Do use action buttons to enable users to select what to display in new tabs." Or something like that.

src/app/components/tabs/index.html
```diff
@@ -1,15 +1,292 @@
...
+					<div class="anatomy-item anatomy-item-required"><div class="anatomy-marker">2</div><div class="anatomy-description">Unselected tab</div></div>
+					<div class="anatomy-item anatomy-item-required"><div class="anatomy-marker">3</div><div class="anatomy-description">Tab border</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">4</div><div class="anatomy-description">Tab count (optional)</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">5</div><div class="anatomy-description">Close tab button (optional)</div></div>
+					<div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">6</div><div class="anatomy-description">Open tab button (optional)</div></div>
+				</div>
+			</div>
+			<div class="two-col-image">
+				<div class="screenshot">
+					<img class="bordered" srcset="~/assets/img/guidelines/tab/anatomy.png 1x, ~/assets/img/guidelines/tab/anatomy_2x.png 2x" src="~/assets/img/guidelines/tab/anatomy.png" />
```
blackbaud-johnly: In this graphic, it looks slightly odd to me that we use small lines with items 2 and 3 but none the others. I defer to your design expertise, but I wonder if we really need the lines here. I think the placement of 2 makes it clear what it is pointing to without the line, and I think it would be more effective to move 3 down to touch the border instead of using a line to connect it to the border (similar to how #2 touches the highlight border in the corresponding image in the Anatomy section of the popover docs).

src/app/components/popover/index.html
```diff
@@ -1,13 +1,161 @@
...
+					</div>
+				</div>
+			</div>
+		</sky-demo-page-content>			
+			
+		<sky-demo-page-content sectionHeading="See more">
```
blackbaud-johnly: What do you guys think of changing this to something like "Related information"? ... Also, should we break these into separate columns? I don't see anything in https://blackbaudinc.app.box.com/s/qfhjf8t3yidcphlkkd7nn440t04n54al/file/402353452509, but I feel like I saw a prototype at some point where the Components section was on the left and the Guidelines section was on the right.

Blackbaud-ToddRoberts:
Added content section components:

app/shared/docs-components:
anatomy-item
two-col-example, two-col-example-text, two-col-example-image

app/components/shared:
component-states-table, component-states-table-item
demo-page-module-info
tabbed-demo-page
updated demo-page-property, demo-page-properties

src/app/components/popover/index.html
Blackbaud-ToddRoberts: yeah, for the most part we'll only need captions for do's/don'ts

src/app/components/tabs/index.html
Blackbaud-ToddRoberts: On the UX side we want to get away from having separate Add and Open buttons, since they are not being used consistently anyway. The desired interaction is to only use the Add button, with action buttons on the new tab view as to what to put into that view.

src/app/components/shared/demo-page-properties.component.html
Blackbaud-ToddRoberts: Good question, I'm not sure. Devs can probably weigh in on that.

src/app/components/popover/index.html
blackbaud-johnly: Sounds good.

src/app/components/shared/demo-page-properties.component.html
blackbaud-johnly: I'll bring this up at the weekly check-in meeting.

src/app/components/tabs/index.html
blackbaud-johnly: A couple things:

First, the Anatomy list and graphic are out of whack. The label for item 6 is "Open tab button," but the graphic has the Add tab button.

Second, recommending only using the Add tab button seems confusing to me. Based on the demos, the Add button creates a new tab in the tabset, while the Open button (seems to) to open an existing tab that was closed. So I don't follow how the Add button is a replacement for the Open button or vice versa. If we are recommending only using the Add button, then how do users reopen tabs after they close them?

Finally, if this is correct and we are recommending that designers only use the Add button, then shouldn't we adjust the description of the `openTab` event that creates the Open button to let them know not to use it? And shouldn't we also add something to the design docs to call out that designers should not use the Open button? Just excluding the button from the Anatomy list and graphic doesn't seem sufficient to make it clear that we want users to use the Add tab button and not use the Open tab button.

Blackbaud-ToddRoberts: Updated the UX guidelines to refer to Add rather than Open. The expected UX for opening a closed tab is in the "Options > Add and close" section. We should update the demo to remove the Open button. Not sure what our deprecation plan is, whether we should deprecate the open button or leave it but recommend against using it.

src/app/components/tabs/index.html
Blackbaud-ToddRoberts: updating the wording around this guideline. I don't think we want a caption since it's not something that should be done in all cases.

src/app/components/tabs/index.html
Blackbaud-ToddRoberts: The general approach we are taking is to only use the pointer lines where necessary, even if it leads to mixed instances with and without. In this case there is some awkwardness of overlaying the dots on the dark blue elements making the pointers necessary on 3.

src/app/components/tabs/index.html
blackbaud-johnly: OK. Thanks.

src/app/components/tabs/index.html
blackbaud-johnly: Makes sense. Thanks.

src/app/components/popover/index.html
Blackbaud-ToddRoberts: updated the heading. Not sure if we want columns in case we run into scenarios where we have related info beyond components and guidelines - seems easier to add another section vertically than another column if that happens.

Blackbaud-ToddRoberts:
@Blackbaud-SteveBrush updated based on your comments if you want to recheck.

Blackbaud-SteveBrush:
@Blackbaud-ToddRoberts This looks good, as long as @blackbaud-johnly and @Blackbaud-DeniseCisneros approve!

Issue #324, opened 2019-02-27
Document how to float controls to the right in toolbar


blackbaud-johnly:
See https://github.com/blackbaud/skyux-layout/pull/18/files. Also https://github.com/blackbaud/skyux-lists/issues/19.

PR #381, opened 2019-04-01
Added child component to toolbar to float controls to the right

Blackbaud-DeniseCisneros:


Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/324

src/app/components/toolbar/index.html
```diff
@@ -20,6 +20,13 @@
     >
     Specifies a section to group items within the toolbar. The section displays items in a separate horizontal row.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="sky-toolbar-view-actions"
+      isOptional="true"
+    >
+    Adds a section on the right side of the toolbar for items that substantially change or affect the view of the toolbar. This includes simple filters and view switchers. If the view section includes more than one item, simple filters appear on the left and view switchers appear on the right.
```
blackbaud-johnly: LGTM, but a couple minor thoughts:

In the first sentence, do we need both "change" and "affect"? Would one of the other (or an alternative such as "alter") work just as well?

In the last sentence, it's probably just me, but I'd delete "view" before "section." The first time I read it, it confused me at first because I thought it sounded like we were talking about a different section somewhere else. I think deleting "view" would make this clearer, but that might just be me.

Also in the last sentence, we say simple filters go on the left and view switchers go on the right as if those are the only two choices. But the previous sentence makes it sound like those are just 2 possible items for this section. Are there other items that can appear in this section? And if so, how does their placement factor in to sections with multiple items?

Blackbaud-TrevorBurch: `Adds a section on the right side of the toolbar for items that alter the view of the toolbar.` is not really true. These items will affect the "view below". The usage guideline state that the items in this section are:
`items that substantially change or affect the view of the rest of the content container.`

Blackbaud-TrevorBurch: So for John's last question - we do not control the ordering here (or on the list version). Consumers give us the items to put in this section and we just render it. 

The usage guidelines do state that the order of common items for this section should be
1. Simple filters
2. Expand all/Collapse all buttons
3. View switcher

From a technical standpoint consumers could put whatever they want in this section and the ordering is totally up to them. This would go against the usage guidelines though.

src/app/components/toolbar/index.html
```diff
@@ -20,6 +20,13 @@
     >
     Specifies a section to group items within the toolbar. The section displays items in a separate horizontal row.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="sky-toolbar-view-actions"
```
Blackbaud-TrevorBurch: This new section is a component not a property. The component will be placed along side your other toolbar items. An example of how this can be used is 


```
<sky-toolbar>
    <sky-toolbar-item>
      <button
        class="sky-btn sky-btn-default"
      >
        Button 1
      </button>
    </sky-toolbar-item>
    <sky-toolbar-item>
      <button
        class="sky-btn sky-btn-default"
      >
        Button 2
      </button>
    </sky-toolbar-item>
    <sky-toolbar-view-actions>
      <button
        class="sky-btn sky-btn-default"
        title="Sort descending"
        type="button"
      >
        <sky-icon
          icon="angle-double-down"
        >
        </sky-icon>
      </button>
      <button
        class="sky-btn sky-btn-default"
        title="Sort ascending"
        type="button"
      >
        <sky-icon
          icon="angle-double-up"
        >
        </sky-icon>
      </button>
    </sky-toolbar-view-actions>
  </sky-toolbar>
```

Blackbaud-TrevorBurch: The list version of this is not a property either. We'll need to fix it there too if that is how it is documented.

Blackbaud-DeniseCisneros: I'll create a separate issue for updating the list version of the property

src/app/components/toolbar/index.html
```diff
@@ -8,6 +8,15 @@
     The toolbar component displays a SKY UX-themed toolbar. Use the <stache-code>sky-toolbar</stache-code> component in conjunction with the <stache-code>sky-toolbar-item</stache-code> child component.
   </sky-demo-page-summary>
 
+  <sky-demo-page-properties sectionHeading="Toolbar component">
```
blackbaud-johnly: I don't think this should be a new section. I think `sky-toolbar-view-actions` is another child component to add to the list in the existing Toolbar child components section. @Blackbaud-TrevorBurch, is that correct, or am I missing something?

Otherwise, this LGTM. If Trevor agrees, I think we just need to move this `sky-demo-page-property` subsection to the other section.

Blackbaud-TrevorBurch: You are right @blackbaud-johnly. That is my mistake. I told @Blackbaud-DeniseCisneros that it was wrong earlier but I misread the stache code.... It should go alongside `sky-toolbar-item` and `sky-toolbar-section` in the `Toolbar child components` section.

Sorry about that @Blackbaud-DeniseCisneros!

Blackbaud-DeniseCisneros: No problem! Fixed :)

Issue #341, opened 2019-03-08
Update Installation section in component docs


blackbaud-johnly:
Add text around code block in new Installation section to provide context and to call out adding the module to `app-extras`. The `app-extras` info needs to address 3 scenarios: SKY UX SPAs, libraries, and non-SKY UX SPAs that are consuming the component.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/383.

PR #383, opened 2019-04-02
Add guidance to Installation section

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/341. Also cleans up a handful of component topics where the module and/or package were not displaying correctly. (There are several others I wasn't able to fix here. I will create a separate issue for those after this update.)

### DIFF
--- a/src/app/components/shared/demo-page.component.html
+++ b/src/app/components/shared/demo-page.component.html
@@ -27,11 +27,11 @@
       Installation
     </stache-page-anchor>
     <p>
-      To use <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span>, you must first install the module's NPM package to set it up as a dependency. In the command prompt, run the following command from your SPA's root directory:
+      To use any exports from <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span>, you must install it as a dependency to your project. In the command prompt, run the following command from your SPA's root directory:
     </p>
     <stache-code-block [code]="npmInstall"></stache-code-block>
     <p>
-      After the module is in place, use an Angular <stache-code>app.module.ts</stache-code> file to make it available to your SPA. For SKY UX SPAs, add <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span> as both an import and an export in the <stache-code>app-extras.module.ts</stache-code> file that controls the rendering of the SPA. For SKY UX libraries, add <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span> as an import in the module for the component that consumes it.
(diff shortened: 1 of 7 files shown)


src/app/components/shared/demo-page.component.html
```diff
@@ -26,7 +26,13 @@ <h4 *ngIf="skyuxModulesForDisplay">
     <stache-page-anchor>
       Installation
     </stache-page-anchor>
+    <p>
+      To use <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span>, you must first install the module's NPM package to set it up as a dependency. In the command prompt, run the following command from your SPA's root directory:
```
Blackbaud-BobbyEarl: I'm wondering if we could clarify that you use exports from the module, not the module itself.  Things like pipes, components, and services. 

Perhaps `To use any exports from the {{ skyuxModulesForDisplay }} module, you must install it as a dependency to your project.  In the command prompt, run the following command in your SPA's root directory:`

src/app/components/shared/demo-page.component.html
```diff
@@ -26,7 +26,13 @@ <h4 *ngIf="skyuxModulesForDisplay">
...
+    </p>
     <stache-code-block [code]="npmInstall"></stache-code-block>
+    <p>
+      After the module is in place, use an Angular <stache-code>app.module.ts</stache-code> file to make it available to your SPA. For SKY UX SPAs, add <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span> as both an import and an export in the <stache-code>app-extras.module.ts</stache-code> file that controls the rendering of the SPA. For SKY UX libraries, add <span *ngFor="let moduleName of skyuxModulesForDisplay"><stache-code>{{ moduleName }}</stache-code></span> as an import in the module for the component that consumes it.
```
Blackbaud-BobbyEarl: I'd like to see if we can make the guidance more generic, but then be explicit about what to do when using our SDK.  It is also only necessary to add it to the `exports` section.  Perhaps something like...

"After installation, you need to make any exports from the module available to your SPA by adding it to the `exports` section of your SPA's main module.  If you are using SKY UX SDK, add it to the exports section of your `app-extras.module.ts` file."

src/app/components/shared/demo-page.component.html
blackbaud-johnly: Makes sense.

src/app/components/shared/demo-page.component.html
blackbaud-johnly: Sounds good.

blackbaud-johnly:
@Blackbaud-BobbyEarl, I made updates to address your feedback. Please just let me know if you see anything else.

Blackbaud-BobbyEarl:
I think getting https://github.com/blackbaud/skyux2-docs/pull/383 in first will simplify some of the nasty logic we have to support in the template, and clean up this PR slightly.

blackbaud-johnly:
Marked as hold pending https://github.com/blackbaud/skyux2-docs/pull/398.

Issue #259, opened 2019-02-08
Document `SkyLibResources` provider and pipe


Blackbaud-SteveBrush:
Source: https://github.com/blackbaud/skyux-i18n/blob/master/src/app/public/modules/i18n/lib-resources.service.ts

Blackbaud-SteveBrush:
### Install the plugin

**Install the SKY UX Builder plugin:**

```
npm install --save-dev --save-exact @skyux-sdk/builder-plugin-skyux
```

**List the plugin in your `skyuxconfig.json` file:**

```
{
  "plugins": [
    "@skyux-sdk/builder-plugin-skyux"
  ]
}
```

### Create a resources provider

Create a new directory named `plugin-resources` at `./src/app/public/plugin-resources` (the name and location of this directory is important!).

Inside the `plugin-resources` directory create a new file named  `my-library-resources-provider.ts` (replace `my-library` with the name of your library).

<img width="521" alt="Screen Shot 2019-04-11 at 10 13 47 AM" src="https://user-images.githubusercontent.com/12497062/55964320-84f8b700-5c42-11e9-8379-a59df1bec8f4.png">

The contents of the `my-library-resources-provider.ts` file should look like this (replace `MyLibrary` with the name of your library):

```
import {
  SkyAppLocaleInfo,
  SkyLibResourcesProvider
} from '@skyux/i18n';

export class MyLibraryResourcesProvider implements SkyLibResourcesProvider {
  public getString: (localeInfo: SkyAppLocaleInfo, name: string) => string;
}

```

### Create a resources module

The resources module will be shared by all modules in your library that need access to locale resources strings.

In a directory that's used for shared types, create a new Angular module with the following contents (replace `MyLibrary` with the name of your library):

<img width="465" alt="Screen Shot 2019-04-11 at 10 17 01 AM" src="https://user-images.githubusercontent.com/12497062/55964626-0b14fd80-5c43-11e9-9f2a-071da050a09a.png">

```
import {
  NgModule
} from '@angular/core';

import {
  SkyI18nModule,
  SKY_LIB_RESOURCES_PROVIDERS
} from '@skyux/i18n';

import {
  MyLibraryResourcesProvider
} from '../../plugin-resources/my-library-resources-provider';

@NgModule({
  providers: [{
    provide: SKY_LIB_RESOURCES_PROVIDERS,
    useClass: MyLibraryResourcesProvider,
    multi: true
  }],
  exports: [
    SkyI18nModule
[...]

PR #408, opened 2019-04-11
Added skyLibResources

Blackbaud-DeniseCisneros:
Added docs around skyLibResources. 

I'm open to suggestions on format. I didn't want to break this into two pages under Localization to avoid multiple subfolders but maybe that makes more sense?

Issue: https://github.com/blackbaud/skyux2-docs/issues/259


Blackbaud-SteveBrush:
@Blackbaud-DeniseCisneros Just some context, the library resources provider (and pipe) are used by libraries, only. Does that change the location of where this should go?

blackbaud-johnly:
Do you have an opinion on where to put this content, @Blackbaud-SteveBrush? If the new content is distinct in that it is specific to libraries but it is still also for localization, maybe it should go in a child topic under the current localization topic. What do you think?

Blackbaud-TrevorBurch:
@blackbaud-johnly I think where you currently have these topics is good with the link that was added from the component libraries topic.

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -163,4 +171,132 @@
...
+    })
+    export class MyLibraryResourcesModule { }
+  </sky-code-block>
+
+  <h3>
+    Additional Setup
+  </h3>
+
+  <p>
+    Import the new <stache-code>MyLibraryResourcesModule</stache-code> into modules requiring access to locale resource strings. This allows the module to use the <stache-code>skyLibResources</stache-code> pipe and the <stache-code>SkyLibResourcesService</stache-code>.
```
Blackbaud-TrevorBurch: This content is somewhat more buried in this topic than for SPAs. Should we call out that localization uses the pipe and service at the top of the library section and then flow into how to setup the library to use them? I feel like that might make sense as after initial setup the pipe and service will be what is used repeatedly.

Blackbaud-DeniseCisneros: I pushed the Localization for libraries section up so that it's before the Localization for SPAs.

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+</ol>
+
+<h3>
+  Create a resources provider
+</h3>
+
+  <ol>
+    <li>
+      <p>
+        Create a new directory named <stache-code>plugin-resources</stache-code> in <stache-code>./src/app/public/plugin-resources</stache-code>. The name and location of the directory are important.
```
blackbaud-johnly: I'd delete "new" because it's redundant. I'd also change the start to "Create a `plugin-resources` directory in..."

Also, I'd delete the period before `/src` in the folder path. We don't usually have a period in front of `/src` in our paths.

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+      </p>
+    </li>
+    <li>
+      <p>
+        Inside the <stache-code>plugin-resources</stache-code> directory, create a new file named <stache-code>my-library-resources-provider.ts</stache-code> and replace <stache-code>my-library</stache-code> with the name of your library. In the sample below, replace <stache-code>MyLibrary</stache-code> with the name of your library.
```
blackbaud-johnly: I'd change "Inside" to "In" because it means the same thing and is shorter.

Also, I'd delete "new" again because it's redundant and I'd rearrange the bit after the opening clause to "create a `my-library-resources-provider.ts` file..." Also in that sentence, I'd change "and replace" to "where" and "with" to "is."

It might juts be me, but it seems like the third sentence should be fleshed out and maybe even made into a separate step. Currently, we tell them to replace something from the code block, but we've never told them to enter the code block in the file they created. It seems like we should have something along the lines of "In the new file, import `SkyAppLocaleInfo` and `SkyLibResourcesProvider` from the `@skyux/i18n` library and export a `MyLibraryResourcesProvider` class where `MyLibrary` is the name of your library." Or something like that.

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+  </sky-code-block>
+
+  </ol>
+
+  <h3>
+    Create a resources module
+  </h3>
+
+  <p>
+    The resources module will be shared by all modules in your library that need access to locale resources strings.
```
blackbaud-johnly: I'd reword this to use active voice: "The resources module allows all the modules in your library to access the locale resource strings."

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+  </p>
+
+  <p>
+    In a shared type directory, create a new Angular module with the content below. Replace <stache-code>MyLibrary</stache-code> with the name of your library.
```
blackbaud-johnly: I assume "type" is the name of the directory. If so, it should be a code snippet. If not, what does "shared type directory" mean? ... I'd also delete "new." ... Also, it's probably just me, but I'd combine the two sentences: "... create the following Angular module where `MyLibrary` is the name of your library."

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+      }],
+      exports: [
+        SkyI18nModule
+      ]
+    })
+    export class MyLibraryResourcesModule { }
+  </sky-code-block>
+
+  <h3>
+    Additional Setup
```
blackbaud-johnly: "setup" should be lowercase

src/app/learn/get-started/advanced/localization/index.html
```diff
@@ -4,18 +4,154 @@
...
+  </h3>
+
+  <p>
+    Import the new <stache-code>MyLibraryResourcesModule</stache-code> into modules requiring access to locale resource strings. This allows the module to use the <stache-code>skyLibResources</stache-code> pipe and the <stache-code>SkyLibResourcesService</stache-code>.
```
blackbaud-johnly: I'd change "the new" to "your." ... I'd add "any" before "modules" and change "requiring" to "that require." ... Also, do you think we should mention that `MyLibrary` is just a placeholder? We could flip the sentence around: "For modules that require access to locale resource strings, import `MyLibraryResourcesModule` where `MyLibrary` is the name of your library."

In the second sentence, "module" should be "modules" to be consistent with the plural "modules" in the first sentence. And I'd either delete "the" before the service name or add "service" after the code snippet.

src/app/learn/get-started/advanced/localization/index.html
Blackbaud-DeniseCisneros: Checked with Bobby on this and he said to remove "type" so I did, and I think it's clearer now.

Issue #258, opened 2019-02-13
Document `SkyCloseModalArgs` for modal component


blackbaud-johnly:
Modal event emits a `SkyCloseModalArgs` with a property called `data`, that's set to whatever object you pass into the Modal instance .save(<object>) function. See https://blackbaud.slack.com/conversation/C3TQ4C0KG/p1550026182060800.

PR #412, opened 2019-04-15
Added values for reason property in skyCloseModalArgs

Blackbaud-DeniseCisneros:
Based on the Slack thread, I took away that we just needed to call out the values. Let me know if we need to make a proper section for this.

Addresses https://github.com/blackbaud/skyux2-docs/issues/258

### DIFF
--- a/src/app/components/modal/index.html
+++ b/src/app/components/modal/index.html
@@ -126,7 +126,7 @@
     <sky-demo-page-property
       propertyName="closed"
     >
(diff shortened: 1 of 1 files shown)


src/app/components/modal/index.html
```diff
@@ -126,7 +126,7 @@
     <sky-demo-page-property
       propertyName="closed"
     >
-    An event that the modal instance emits when it closes. It emits a <stache-code>SkyCloseModalArgs</stache-code> object with a <stache-code>data</stache-code> property that includes data passed from users on close or save and a <stache-code>reason</stache-code> property that indicates whether the modal was saved or closed without saving.
+    An event that the modal instance emits when it closes. It emits a <stache-code>SkyCloseModalArgs</stache-code> object with a <stache-code>data</stache-code> property that includes data passed from users on close or save and a <stache-code>reason</stache-code> property that indicates whether the modal was saved or closed without saving. The <stache-code>reason</stache-code> property accepts string values such as <stache-code>cancel</stache-code>, <stache-code>close</stache-code>, and <stache-code>save</stache-code>.
```
Blackbaud-BobbyEarl: I wonder if would could emphasize it's a string and the examples you listed are examples.  Something like
```
The "reason" property accepts any string value.  Common examples are "cancel," "close," and "save."
```

Issue #415, opened 2019-04-17
List Filters > package is wrong


Blackbaud-AlexKingman:
https://developer.blackbaud.com/skyux/components/list/filters

Defined in `SkyListFiltersModule` in NPM package `@skyux/lists`

... should be ...

Defined in `SkyListFiltersModule` in NPM package `@skyux/list-builder`

Blackbaud-DeniseCisneros:
Fixed in https://github.com/blackbaud/skyux2-docs/pull/416

Issue #407, opened 2019-04-11
Update Installation for iOS Simulator on Mac


Blackbaud-DeniseCisneros:
Reported by Christi Schneider

Step 3 is different on iOS 12 (not sure when this changed)
When you drag the cert onto the sim, you get an alert saying “This website is trying to download a configuration profile. Do you want to allow this?” and you should press Allow. Then it has another alert that says “Profile Downloaded. Review the profile in Settings app if you want to install it. Close”

You have to open the Settings app, General, Profile, SKY UX, click Install 3 times (edited)
The rest is the same, just an extra step now.

PR #416, opened 2019-04-17
Installation edit

Blackbaud-DeniseCisneros:
@Blackbaud-AlexKingman Updated the package name

@Blackbaud-ChristiSchneider If you could confirm the step I added is correct, that would be awesome :)

Addresses https://github.com/blackbaud/skyux2-docs/issues/407 and https://github.com/blackbaud/skyux2-docs/issues/415

### DIFF
--- a/src/app/learn/get-started/prereqs/ssl-certificate/index.html
+++ b/src/app/learn/get-started/prereqs/ssl-certificate/index.html
@@ -78,6 +78,11 @@
     </li>
     <li>
       <p>
+        Open the <strong>Settings</strong> app and navigate to <strong>General</strong> > <strong>Profile</strong> > <strong>SKY UX</strong>.
+      </p>
+    </li>
+    <li>
+      <p>
         Select <strong>Install</strong> on each page (3 times total) to install the certificate.
       </p>
     </li>
--- a/src/app/learn/get-started/prereqs/ssl-certificate/index.html
+++ b/src/app/learn/get-started/prereqs/ssl-certificate/index.html
@@ -73,7 +73,7 @@
     </li>
     <li>
       <p>
-        In the prompt that appears, select <strong>Allow</strong> and then <strong>Close</strong>.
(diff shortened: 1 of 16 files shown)


src/app/learn/get-started/prereqs/ssl-certificate/index.html
```diff
@@ -76,6 +76,11 @@
         In the prompt that appears, select <strong>Allow</strong> to open a settings page.
```
Blackbaud-ChristiSchneider: This no longer opens the settings page, instead, it shows an alert that says "Profile Downloaded. Review the profile in Settings app if you want to install it.", and the only option is "Close".

So this line could say "... select <strong>Allow</strong> and then <strong>Close</strong>" and then I think it's good!

Blackbaud-ChristiSchneider:
Looks good!

src/app/learn/get-started/prereqs/ssl-certificate/index.html
```diff
@@ -73,7 +73,12 @@
     </li>
     <li>
       <p>
-        In the prompt that appears, select <strong>Allow</strong> to open a settings page.
+        In the prompt that appears, select <strong>Allow</strong> and then <strong>Close </strong>.
```
blackbaud-johnly: What do you think of adding "to download the profile" after "Allow" and adding "select" before "Close"? No big deal either way; I just thought that might add some context about what's happening between the selections.

src/app/learn/get-started/prereqs/ssl-certificate/index.html
```diff
@@ -73,7 +73,12 @@
     </li>
     <li>
       <p>
-        In the prompt that appears, select <strong>Allow</strong> to open a settings page.
+        In the prompt that appears, select <strong>Allow</strong> and then <strong>Close</strong>.
```
blackbaud-johnly: What do you think of adding "to download the profile" after "Allow" and adding "select" before "Close"? No big deal either way; I just thought that might add some context about what's happening between the selections.

Issue #420, opened 2019-04-18
Grids > update `sortField` property


Blackbaud-AlexKingman:
The `sortField` property is NOT inteded to sort data. It's just for showing the visual indicator (caret). I'd suggest we update our docs to read something like this (needs tweaking):

`sortField` — (Optional.) Allows consumer to programmatically control the visual indicator of a sorted field. This is useful if a consumer wants to programmatically sort the data, but also tell the grid column to display the caret. This property accepts a `ListSortFieldSelectorModel` value that contains a `fieldSelector` string to represent the current sort field and a descending `boolean` to indicate which direction the caret should display.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/424.

PR #424, opened 2019-04-19
Update sortField property and break out ListSortFieldSelectorModel docs

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/420.

### DIFF
--- a/src/app/components/grid/index.html
+++ b/src/app/components/grid/index.html
@@ -86,7 +86,7 @@
       propertyName="sortField"
       isOptional="true"
     >
-    Displays a caret in the column that sorts the grid. This is particularly useful when you programmatically sort data and want to visually indicate how you sort the grid. This property accepts <a stacheRouterLink="." fragment="listsortfieldselectormodel-properties">a <stache-code>ListSortFieldSelectorModel</stache-code> value</a> to represent the current sort field and the sort order.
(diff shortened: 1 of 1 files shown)


src/app/components/grid/index.html
```diff
@@ -86,7 +86,7 @@
       propertyName="sortField"
       isOptional="true"
     >
-    Specifies the field to use to sort the grid. This property accepts a <stache-code>ListSortFieldSelectorModel</stache-code> value that contains a <stache-code>fieldSelector</stache-code> string to represent the current sort field and a <stache-code>descending</stache-code> boolean to indicate whether to sort in descending order.
+    Displays a caret in the column that sorts the grid. This is particularly useful when you programmatically sort data and want to visually indicate how you sort the grid. This property accepts <a stacheRouterLink="." fragment="listsortfieldselectormodel-properties">a <stache-code>ListSortFieldSelectorModel</stache-code> value</a> to represent the current sort field and the sort order.
```
Blackbaud-AlexKingman: "...that sorts the grid" makes me think its actually doing the sorting. Could this say something like "...that indicates which column has been sorted" ?

blackbaud-johnly: I tweaked this to address the feedback. Let me know if my change works.

PR #433, opened 2019-04-25
inline form demo + docs updates

Blackbaud-ToddRoberts:


Blackbaud-MattGregg:
@Blackbaud-ToddRoberts  One thing I noticed is the example images don’t follow guidelines for button labels. The “Add a price point” includes an article.

Blackbaud-MattGregg:
@Blackbaud-AlexKingman Did this component pass all of its unit accessibility test rules? None of the edit buttons have a text equivalent. Is this just an issue with the demo or is this inherently a problem with the component?

Blackbaud-AlexKingman:
@Blackbaud-MattGregg It did indeed pass. Are you talking about the buttons in the form's footer? Or the little pencil icons? If it's the latter, those pencil icons are only part of the demo and not the component itself. We should probably update those in the demo code if they do not have a text equivalent.

src/app/components/inline-form/index.html
```diff
@@ -151,6 +151,30 @@
 		<header><div class="sky-subsection-heading sky-margin-stacked-default">Edit form</div></header>
 			<p>When users click the Edit button, the view content fades out and the form slides in to replace the view content.</p>
 
+  <header><div class="sky-section-heading">Content</div></header>
+    <header><div class="sky-subsection-heading sky-margin-stacked-default">Primary buttons</div></header>
+  <p>The primary action on the inline form should be Save or Done.</p>
+    <stache-row>
+      <stache-column screenSmall="6">
+        <p>Use Save if the data is saved to the database immediately.</p>
```
blackbaud-johnly: Not a big deal, but I'd tweak this to use active voice: "Use Save if the inline form saves data to the database immediately."

src/app/components/inline-form/index.html
```diff
@@ -151,6 +151,30 @@
...
+      </stache-column>
+      <stache-column screenSmall="6">
+        <div class="screenshot screenshot-success">
+          <img src="~/assets/img/guidelines/inline-form/usage-do-block.png" style="width: 100%;" />
+        </div>
+      </stache-column>
+    </stache-row>
+    <stache-row>
+      <stache-column screenSmall="6">
+        <p>Use Done if the inline form is part of a larger form and the data doesn't get saved until the larger form is saved.</p>
```
blackbaud-johnly: Again, not a big deal, but I'd tweak this to use active voice and rearrange a bit: "Use Done if the inline form does not save data to the database immediately because it is part of a larger form."

Blackbaud-ToddRoberts:
@Blackbaud-MattGregg updated images to follow button text guidelines

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: updated

src/app/components/inline-form/index.html
Blackbaud-ToddRoberts: updated

Issue #404, opened 2019-04-10
Update UA guidelines for empty-state call to action


blackbaud-johnly:
Update the guidelines at https://developer.blackbaud.com/skyux/design/guidelines/user-assistance#empty-state-help. Call out that the best case is for the button to open an inline experience in a modal. If that's not possible and users need to be sent elsewhere, a link should be used instead to imply navigation. The preferred option is for the experience to happen inline, and the button that supports that. But when that's not possible, a link should help indicate navigation where as the button suggests a modal.

PR #435, opened 2019-04-25
Added UA guidelines for empty-state help

Blackbaud-DeniseCisneros:
Let me know if I need to elaborate more.

Addresses https://github.com/blackbaud/skyux2-docs/issues/404

### DIFF
--- a/src/app/design/guidelines/user-assistance/index.html
+++ b/src/app/design/guidelines/user-assistance/index.html
@@ -30,9 +30,6 @@
 				Empty-state help provides user assistance when content containers are not populated. The content can include a title, body copy, and buttons to take actions or access additional resources. The help content does not appear when content containers are populated.
 			</p>
 			<p>
-				When possible, use a button to open an inline experience in a modal. If you can't use a button or if users should be sent elsewhere, use a link to imply navigation. Buttons suggest modals while links indicate navigation. 
-			</p>
-			<p>
 				Empty-state help content explains how users can take advantage of the functionality in question, and it follows these guidelines:
 			</p>
 			<ul>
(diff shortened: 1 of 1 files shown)


src/app/design/guidelines/user-assistance/index.html
```diff
@@ -29,6 +29,9 @@
 			<p>
 				Empty-state help provides user assistance when content containers are not populated. The content can include a title, body copy, and buttons to take actions or access additional resources. The help content does not appear when content containers are populated.
 			</p>
+			<p>
+				When possible, use a button to open an inline experience in a modal. If you can't use a button or if users should be sent elsewhere, use a link to imply navigation. Buttons suggest modals while links indicate navigation.
```
blackbaud-johnly: The placement of this content feels a off to me. I feel like it breaks up the flow of the existing content. I'd make this a the fourth bullet in the unordered list (before the last bullet). Since the second and third bullets pertain to the title and body copy, I think it makes sense to put guidelines for bullets next so that we follow the same order that we use in the first paragraph.

To make that work, I think we'd need to modify your existing text just a bit, and I also think we should update this to reference the button styles that we recommend. I'm thinking something like: "For the button, use the secondary button class to open a modal where users can complete the experience inline. If users cannot complete the experience inline and must navigate elsewhere, use the link button class instead. Secondary buttons indicate modals, while links indicate navigation." ... I'd make "the secondary button class" and "the link button class" both links to https://developer.blackbaud.com/skyux/components/button. ... What do you think?

Blackbaud-DeniseCisneros: I like that and updated the PR with this.

Blackbaud-DeniseCisneros:
🤦‍♀ Thank you for catching link edits.

Issue #384, opened 2019-04-03
Fix component pages that are missing module and/or package info


blackbaud-johnly:
A handful of the docs for component are missing the module and/or package info at the top of the page. (When the package info is missing, the Installation section is missing.) I fixed a handful of these while updating the Installation section, but a few required some extra effort. They include button, form, navbar, page summary, status indicator, and text expand.

blackbaud-johnly:
Also, follow up on note from old issue that says "Docs for List view grid but it doesn't say anything about which module contains sky-list-column-selector-action."

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/443.

PR #443, opened 2019-05-01
Update packages

Blackbaud-DeniseCisneros:
Added missing package info thanks to @Blackbaud-TrevorBurch's fix

Addresses https://github.com/blackbaud/skyux2-docs/issues/384

### DIFF
--- a/src/app/components/page-summary/index.html
+++ b/src/app/components/page-summary/index.html
@@ -16,59 +16,59 @@
     </sky-alert>
   </sky-demo-page-summary>
 
-  <sky-demo-page-content sectionHeading="Title and subtitle">
+  <sky-demo-page-properties sectionHeading="Title and subtitle">
   <p>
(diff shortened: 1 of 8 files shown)


src/app/components/page-summary/index.html
```diff
@@ -18,75 +16,63 @@
     </sky-alert>
   </sky-demo-page-summary>
 
-  <stache-page-anchor>
-    Title and subtitle
-  </stache-page-anchor>
-
+  <sky-demo-page-content sectionHeading="Title and subtitle">
```
blackbaud-johnly: We need to change the `sky-demo-page-content` sections to `sky-demo-page-properties` sections. The `sky-demo-page-content` sections always appear at the end of the page, which is why UX guidelines are always last. This didn't matter on the button or form docs because those docs don't have demos, but the default behavior of the `sky-demo-page-content` sections messes things up a bit here. Changing these sections to use `sky-demo-page-properties` resolves the placement issue, and since all the sections just include a couple `p` tags, I don't think it messes anything else up. .... We should also bring this up with the full team to make sure there are no issues that I'm overlooking with this workaround and to see if we want devote engineering resources to working on a property for `sky-demo-page-content` that provides flexibility in where to place those sections.

PR #461, opened 2019-05-10
Edit Builder plugin docs

blackbaud-johnly:


src/app/learn/reference/extensibility/index.html
```diff
@@ -16,80 +16,96 @@
...
-```
+SKY UX Builder is intended to be listed as a `devDependency` in the `package.json` file for every SPA. Builder handles the majority of `skyux` commands, including testing, serving, and building, and it wraps up the logic and configuration for other tools such as the Angular Compiler, Webpack, Karma, and Protractor.
+
+## SKY UX CLI plugins
+
+SKY UX CLI is extensible. Follow these guidelines to create custom commands and modules to handle those commands:
+
+<ul>
+  <li>
+    Install the module globally or locally.
```
Blackbaud-BobbyEarl: Changing the tense of this sentence highlights that this step is really geared towards the consumers.  I'm not sure if it's best to just remove it from here, or to create a new section.  This would be necessary to consume the plugin, but not required for someone creating a plugin.

src/app/learn/reference/extensibility/index.html
```diff
@@ -16,80 +16,96 @@
...
+  </li>
+  <li>
+    Start the module name with `@skyux-sdk/builder`.
```
Blackbaud-BobbyEarl: We need both cases here, but can be more explicit about when to use which one.  "If on the SKY UX team, the module name should following the syntax `@skyux-sdk/builder-[your-module-name-here]`.  For anyone else, use the syntax 'skyux-builder-[your-module-here]`.  The reason for this, is non-SKYUX creators won't have access to publish to the `@skyux-sdk/` scope.

src/app/learn/reference/extensibility/index.html
```diff
@@ -16,80 +16,96 @@
...
+          }
+        }
+      }
+    </sky-code-block>
+  </li>
+</ul>
+
+## SKY UX Builder plugins
+
+SKY UX Builder is extensible. Follow these guidelines to create custom packages:
```
Blackbaud-BobbyEarl: I like the addition of these sentence, but maybe we could add how it's extensible.  "SKY UX Builder is extensible but in a different way.  Each file that's parsed as part of your project can be modified at build time.  This technique can be very useful for dynamically injecting common boilerplate behavior."

src/app/learn/reference/extensibility/index.html
```diff
@@ -16,80 +16,96 @@
...
+
+<ul>
+  <li>
+    Add the package as a dependency in `package.json`.
```
Blackbaud-BobbyEarl: Let's wrap this in a `<p>` instead of the two `<br>`s.

Blackbaud-BobbyEarl: Similar to my feedback for the CLI section, this step and the next one are really geared for consumers.  I'm not sure how best to differentiate that though.

Issue #401, opened 2019-04-09
Document design guidelines for page background


blackbaud-johnly:
Document approved changes related to our design strategy to use white backgrounds on all pages and only apply gray to the `tile-dashboard` component. See https://blackbaud.invisionapp.com/share/XSQDELX2MFZ.  

Implement this guideline as the default behavior is a separate task, but we want to publish the design guidelines and update site on 4 pages:
- Components > Tile
- Design > Guidelines > Page design
- Design > Guidelines > Content containers
- Design > Styles > Colors

See @Blackbaud-AdamFunderburk for more details.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/505 and https://github.com/blackbaud/skyux2-docs/pull/477.

PR #477, opened 2019-05-14
Added design guidelines for multiple components

Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/401

Only item still pending from issue is the addition of `sky-background-color-page-default`


Blackbaud-AdamFunderburk:
Small typo on the "Design > Guidelines > Content containers" page:

In the first sentence for "Page section" - "Pages can be divided into vertical section**[s]**..."

src/app/components/tile/index.html
```diff
@@ -175,4 +175,57 @@
...
+    </stache-column>
+    <stache-column screenSmall="6">
+      <div class="screenshot screenshot-danger"><img src="~/assets/img/guidelines/tile/bad-example-1.png" style="width: 100%;" /></div>
+    </stache-column>
+  </stache-row>
+
+  <stache-row>
+    <stache-column screenSmall="6">
+      <p>
+        Don't use tiles as a container for all content on a page outside of the summary.
```
blackbaud-johnly: I wonder if we really need this guideline. I mean, isn't it implicit from the previous guideline? No harm in including it, I guess, but I wonder if it's worth asking Todd if we really want this guideline.  

I think this might be a scenario where we want to use singular instead of plural. What do you think of changing this to something like "Don't use a single tile to contain all of a page's content after the summary."?

src/app/design/guidelines/content-containers/index.html
```diff
@@ -7,15 +7,29 @@
     Divide a page into sections of related content
```
blackbaud-johnly: I know this is existing content, but it might be worth double-checking with Todd to see if tiles are only for pages. That's how everything is written here, but I assume tiles can also be used in modals.

src/app/design/guidelines/page-layouts/index.html
```diff
@@ -90,6 +90,17 @@
     The following sections describe the standard SKY UX page layouts. These layouts show how to package components to suit a variety of user needs, but the system can accommodate alternate layouts as necessary.
   </p>
 
+  <div class="sky-subsection-heading" style="margin-top: 20px;">
+    Sectioned page
+  </div>
+
+  <p>
+    A sectioned page is a simple layout that defines page sections with typography and spacing. Use vertical page sections to categorize the relevant content on the page and group. Use <a routerLink="/components/fluid-grid">fluid grid</a> to arrange content horizontally. The page can use most content containers like page summary, tabs, repeaters, cards, etc. Keep the number of sections low (5 or fewer), or consider breaking the content up with the additional tabs or the tile-based Work center layout.
```
blackbaud-johnly: I don't understand what the second sentence means. What does "to categorize the relevant content on the page and group" mean? And why does it only apply to vertical sections? Also, the content container topic makes it sound like you use typography and spacing for vertical page sections OR you use fluid grid for horizontal page sections. But this first sentence makes it sounds like you use typography and spacing for all page sections, which suggests that you use typography, spacing, AND fluid grid for horizontal sections. Which is correct? ... I suspect we should change these opening sentences to something like "The sectioned page layout provides a simple way to organize page contents into categories and groups. You can divide pages into vertical sections with typography and spacing, and you can divide pages into horizontal sections with the fluid grid component." But I'm not 100 percent sure if that's an accurate assessment of how this works.

In the third sentence, I'd change "The page" to "Sectioned pages." I'd also change "like" to "such as" and delete "etc." and add "and" before "cards."

In the last sentence, I'd change the opening clause to "Limit the number of sections to 5 or fewer." And I'd make the second half of the sentence a new sentence that makes it clearer that if you need more than 5 sections, you should try something different: "If you need more than 5 sections, break content up tabs or the tile-based work center layout."

src/app/design/guidelines/content-containers/index.html
Blackbaud-DeniseCisneros: Checked in with Todd about this. He said you can use tiles in modals, but ideally an alternate approach would be used instead.

src/app/components/tile/index.html
```diff
@@ -175,4 +175,57 @@
...
+    </stache-column>
+    <stache-column screenSmall="6">
+      <div class="screenshot screenshot-success"><img src="~/assets/img/guidelines/tile/tile-example.png" style="width: 100%;" /></div>
+    </stache-column>
+  </stache-row>
+
+  <stache-row>
+    <stache-column screenSmall="6">
+      <p>
+        Don't use tiles to divide a page into a small number of sections.
```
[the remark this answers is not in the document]
Blackbaud-DeniseCisneros: I swapped out "a small number of" for "unnecessary," but I don't know if that's clear enough even with the example image.

src/app/design/guidelines/content-containers/index.html
blackbaud-johnly: OK. Sounds like we should leave the references to "page" only throughout. Thanks for checking!

src/app/components/tile/index.html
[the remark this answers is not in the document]
blackbaud-johnly: I think I'd stick with "a small number." The wording is awkward, but it's more precise than "unnecessary," which could be interpreted to mean something else. ... Actually, now that I'm looking at this, I think maybe the problem isn't the wording but the guideline itself. I mean, what is "a small number"? Is it 3? 4? 5? I assume this is deliberately vague, but it would be more useful and clear if we set a number. Do you think it's worth reaching out to Todd to see if we can set a number? If he can set a number, this guideline could be "Don't use tiles to divide pages into fewer than XXX sections."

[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: It's hard to define an exact number, since the utility of the behaviors of Tiles rises with the size/complexity of each category, not just the number of categories. In other words, you probably need to exceed 4 or 5 categories (of simple details with a few fields each) before Tiles become appropriate, but you might only need 3 categories when they're deep/rich enough for different users to _repeatedly and reliably_ prioritize one over the others.

Thoughts on how to make that point?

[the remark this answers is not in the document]
blackbaud-johnly: Maybe we can set 5 as the recommendation and then qualify it. What do you think of something along the lines of:
"Don't use tiles to divide pages into fewer than 5 sections. However, if tile content is complex enough for different users to repeatedly and reliably prioritize fewer tiles, you can divide pages into as few as 3 tiles."
Or:
"Don't use tiles to divide pages into fewer than 5 basic categories with simple details or fewer than 3 complex categories with content that is complex enough for different users to repeatedly and reliably prioritize the tiles."

src/app/components/tile/index.html
```diff
@@ -175,4 +175,56 @@
...
+      Tiles are containers to organize varied content. They divide pages into multiple categories and group related content. Use tiles to provide access to a wide range of content where users benefit from the ability to reorder, expand and collapse, and personalize tiles. For more details, see <a routerLink="/design/guidelines/content-containers">content container guidelines</a>.
+    </p>
+
+  </sky-demo-page-content>
+
+  <stache-row>
+    <stache-column screenSmall="6">
+      <div class="sky-subsection-heading">Usage</div>
+      <p>
+        Use tiles to divide pages with many categories of content, especially when the format or presentation differs across categories.
```
Blackbaud-ToddRoberts: The other major factor here is whether the contents need to be reordered. @Blackbaud-AdamFunderburk is the guideline here to use tiles for differing presentations even if there is no need for the user to rearrange? I've generally viewed the need to rearrange as the first check for tiles vs page sections.

src/app/components/tile/index.html
```diff
@@ -175,4 +175,56 @@
     </sky-demo-page-code>
   </sky-demo-page-example>
 
+  <sky-demo-page-content sectionHeading="UX guidelines">
+    <p>
+      Tiles are containers to organize varied content. They divide pages into multiple categories and group related content. Use tiles to provide access to a wide range of content where users benefit from the ability to reorder, expand and collapse, and personalize tiles. For more details, see <a routerLink="/design/guidelines/content-containers">content container guidelines</a>.
```
Blackbaud-ToddRoberts: Should the link to the content container guidelines be in a "Related information" section at the bottom of the page rather than inline here?

Blackbaud-AdamFunderburk: That might be in our new content guidelines, I honestly can't remember. I have no strong argument for that link to go in either place.

src/app/components/tile/index.html
Blackbaud-DeniseCisneros: Checked in with Todd about this, and he said it's worth including because even though it may be redundant, the misuse is common.

src/app/components/tile/index.html
Blackbaud-DeniseCisneros: I'll create that Related guidelines sections

src/app/components/tile/index.html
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: The second option above is a pretty complex sentence. I'll take Door #1.

src/app/components/tile/index.html
Blackbaud-AdamFunderburk: Yeah, the mention of differing presentations might be a red herring here. It's more important to make the point of
"... especially when the content is varied enough that users would benefit from the capability to reorder or expand/collapse the categories."

src/app/components/tile/index.html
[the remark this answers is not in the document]
Blackbaud-DeniseCisneros: Done ✔️

src/app/components/tile/index.html
Blackbaud-DeniseCisneros: Updated with your verbiage, Adam

PR #491, opened 2019-05-23
Doc updates for phone field docs

blackbaud-johnly:


src/app/components/demo-components.service.ts
```diff
@@ -382,7 +382,7 @@ export class SkyDemoComponentsService {
         name: 'Phone field',
         icon: 'phone',
         // tslint:disable-next-line
-        summary: `The phone field module allows users to validate international phone numbers.`,
+        summary: `The phone field module creates a field and dropdown to validate international phone numbers.`,
```
Blackbaud-TrevorBurch: This will not be true with the UI updates that are being made. We can chat around what we want the verbiage here to be.

src/app/components/phone-field/index.html
```diff
@@ -6,7 +6,7 @@
 
   <sky-demo-page-summary>
     <p>
-      The phone field module contains a directive which validates phone numbers against the standards for different countries. The value that is validated against is driven through the <stache-code>ngModel</stache-code> attribute specified on the <stache-code>input</stache-code> element or a <stache-code>FormControl</stache-code> within a reactive form.
+      The phone field module creates a field and dropdown to validate international phone numbers. The phone field input directive validates the phone number that users enter in the field against standards for the country that they select in the dropdown. For validation, phone numbers are driven through the <stache-code>ngModel</stache-code> attribute that you specify on an <stache-code>input</stache-code> element or a <stache-code>FormControl</stache-code> in a reactive form.
```
Blackbaud-TrevorBurch: Same concerns on the `dropdown` language as that is changing.

src/app/components/phone-field/index.html
```diff
@@ -15,9 +15,10 @@
   >
     <sky-demo-page-property
         propertyName="defaultCountry"
+        defaultValue="US"
         [isOptional]="true"
       >
-        Specifies the default country for the phone field input and dropdown. This property accepts <stache-code>string</stache-code> <a href="https://www.nationsonline.org/oneworld/country_code_list.htm">International Organization for Standardization Alpha 2</a> values.
+        Specifies the <a href="https://www.nationsonline.org/oneworld/country_code_list.htm">International Organization for Standardization Alpha 2</a> country code for the default country in the phone field dropdown. This property accepts <stache-code>string</stache-code> values.
```
Blackbaud-TrevorBurch: Same dropdown concerns (both with the old and new versions here).

src/app/components/phone-field/index.html
```diff
@@ -27,18 +28,18 @@
...
-      [defaultValue]="false"
+      defaultValue="false"
       [isOptional]="true"
     >
       Indicates whether to disable the phone field. This property accepts <stache-code>boolean</stache-code> values.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="skyPhoneFieldNoValidate"
-      [defaultValue]="false"
+      defaultValue="false"
```
Blackbaud-TrevorBurch: Good catch. Sorry about that!

src/app/demos/phone-field/phone-field-demo.component.html
```diff
@@ -23,7 +23,7 @@
 <h2
   class="sky-section-heading"
 >
-  Reactive phone field with default country
+  Reactive phone field with a different default country
```
Blackbaud-TrevorBurch: Not sure on the reason for the change here? The other two demos don't specify a default country. Is this due to the fact that the "default" default country is the US? If so, I think maybe should change this to something like "Reactive phone field with a specified default country" or something different. I don't love the "different default' as if I didn't read the docs for the property above (or just missed the US part) then I wouldn't understand what is different about it.

blackbaud-johnly: Yeah, my concern was that the first field was using the default-default country. So labeling this one as having a default country seemed like a weird distinction. I'm not in love with "different default" either. I'll change "different" to "specified."

blackbaud-johnly: Actually, I changed this to "Reactive phone field with a default country specified." Hopefully that works for you.

src/app/components/demo-components.service.ts
blackbaud-johnly: I'm changing "creates a field and dropdown" to "creates a button and text input" both here and in the intro on the the component topic. (I'm also making additional edits to the intro and the rest of the topic.)

src/app/components/phone-field/index.html
blackbaud-johnly: I'm reworking this paragraph.

src/app/components/phone-field/index.html
```diff
@@ -15,9 +15,10 @@
   >
     <sky-demo-page-property
         propertyName="defaultCountry"
+        defaultValue="US"
```
Blackbaud-TrevorBurch: Technically this would be `us` in lower case. I did just push a change to my current changes that allows for capitalized values but we should probably change this just to be safe.

Blackbaud-TrevorBurch: Actually, I just made the change myself :)

PR #509, opened 2019-06-04
fix screenshots

Blackbaud-ToddRoberts:
check:

Homepage
Components > Flyout
Components > Popover
Components > Tabs
Design > Guidelines > Manage Records


Blackbaud-ToddRoberts:
Homepage, Flyout, Manage records should have the old screenshot styling with the 30px colored border. 

Popover, Tabs should have the new screenshot styling without the wide border, and the status indicator caption with a 4px top border (only when there is a caption)

Issue #551, opened 2019-06-21
Document the --no-lint option


Blackbaud-ColbyWhite:
`skyux watch` supports `--no-lint`. it's not [documented](https://developer.blackbaud.com/skyux/learn/reference/cli-commands/watch).

I'm not sure if other commands accept that option.

blackbaud-johnly:
@Blackbaud-BobbyEarl, are there other commands that accept that flag? Like maybe `test`? ... Also, I assume the description for this flag should be something along the lines of "Prevents the command from validating the TypeScript code in the files in the src folder against the rules specified by the tslint.json file in the SKY UX Builder repo." Does that sound about right?

blackbaud-johnly:
Oh, and thanks for filing this, @Blackbaud-ColbyWhite!

Blackbaud-BobbyEarl:
Technically, `build-public-library`, `lint`, `test`, `watch`, and `pact` all reference [`ts-linter` utility class](https://github.com/blackbaud/skyux-sdk-builder/blob/b9db9dbc458f648a7c9c747548e6af350745a39b/cli/utils/ts-linter.js).

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/552.

PR #552, opened 2019-06-24
Document --no-lint flag

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/551.


src/stache/includes/cli/no-lint.html
```diff
@@ -0,0 +1 @@
+<sky-code>--no-lint</sky-code> — Prevents the <a href="https://palantir.github.io/tslint/">TSLint static analysis tool</a> from validating TypeScript code in the <sky-code>src</sky-code> folder against the rules specified by <a href="https://github.com/blackbaud/skyux-builder/blob/master/tslint.json">the <sky-code>tslint.json</sky-code> file in the SKY UX Builder repo</a>.
```
Blackbaud-BobbyEarl: Technically this is only true if the `tslint.json` file in the project's root references builder in it's `extends` property, as seen in the template.  https://github.com/blackbaud/skyux-sdk-template/blob/master/tslint.json#L2

I think we should just say based on your projects `tslint.json` file.

blackbaud-johnly: I updated this and also similar text at https://developer.blackbaud.com/skyux/learn/reference/cli-commands/lint.

PR #555, opened 2019-06-26
add popover trigger videos

Blackbaud-ToddRoberts:


Blackbaud-AdamFunderburk:
The videos are playing well. The frame rate is good, and "click" effect makes it much easier to interpret what's going on.

What's defining the black gradient overlay for the playback controls at the bottom of each preview? It's dense enough relative to our visual design that it risks being distracting, but if that's how our embed player works, then /shrug. It's height appears to be a little unpredictable - the second example has a much bigger gradient overlay than it needs. Can we minimize its height to the playback controls themselves, and not (what appears to be) a percentage of the video height?

Blackbaud-ToddRoberts:
It's the browser default video player. As far as I know we don't have control over the appearance of the controls.

src/app/components/popover/index.html
```diff
@@ -48,14 +48,30 @@ <h3>Positioning</h3>
...
+        </sky-docs-visual-example>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>
+              You can also invoke popovers programmatically. For example, you can programmatically invoke a popover when users visit a page for the first time to highlight an important interface element that they might overlook. However, don't invoke multiple popovers on a page because this reduces their effectiveness.
+            </p>
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <video class="bordered" controls>
+              <source src="~/assets/img/guidelines/popover/popover-auto.mp4" type="video/mp4" />
```
blackbaud-johnly: Is it intentional that this video has the `bordered` class but the previous one does not? When I serve locally, the previous video has a clearly defined, light-grey border that looks the video is smaller than it's container and with a drop shadow within that container. With this video, the image comes right to the edge of its container, with now light-grey border and now drop shadow. Are these differences intentional?

Blackbaud-ToddRoberts: It is intentional, but looking at it we can probably update the first video to not use the shadowed border (which is part of the UI in the video). The intent was for it to look more like a modal, but it's not really needed in this context.

PR #561, opened 2019-07-01
Split view docs

Blackbaud-AlexKingman:


src/app/components/split-view/index.html
```diff
@@ -0,0 +1,229 @@
...
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <img src="~/assets/img/guidelines/split-view/option-list-type-grid.png" />
+            <div class="caption caption-warning sky-text-warning">Use caution when using grids as the list type as they do not perform well at narrow widths.</div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+      <h3>List manipulation</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>You may include a list toolbar so that users can filter, sort and search the list, or add new items to the list.</p>
```
Blackbaud-DeniseCisneros: We've been linking to the component pages as they're called out so I think we should do the same for the list toolbar.

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,229 @@
...
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <img src="~/assets/img/guidelines/split-view/behavior-resize.png" />
+          </sky-docs-visual-example-image>
+      </sky-docs-visual-example>
+    </sky-demo-page-content>
+    <h3>Responsiveness</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>In smaller viewports the split view automatically switches from a side-by-side view to a panel-switching interaction where the user flips between the list and the workspace. The responsive workspace header should have iterator buttons so the user can move between items without having to go back to the list.</p>
```
Blackbaud-DeniseCisneros: Add a comma after viewports?

src/app/components/split-view/index.html
blackbaud-johnly: I created a branch to make some edits; I'll make this edit in my branch.

src/app/components/split-view/index.html
blackbaud-johnly: I created a branch to make some edits; I'll make this edit in my branch.

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,229 @@
...
+    tabHeading="Design">
+
+    <sky-demo-page-content sectionHeading="Usage">
+      <h3 class="sky-text-success">Use when</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>Use a split view if a user needs to work through a list of items with the intent of taking an action on each item in the list.</p>
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <img src="~/assets/img/guidelines/split-view/use-when.png" />
```
blackbaud-johnly: @Blackbaud-ToddRoberts, can we resize this image so that it fits alongside the text? I think it would still be readable at that size. ... Also, shouldn't this image include a caption? I think that's the pattern for images that illustrate when to use or not use components, so it seems like it should have a caption here. Plus, the caption could call out that in this example, checking people in is the action, which would be good because I think it's easy to overlook the check in button.

blackbaud-johnly: Actually, can we just make a new, narrower image instead of resizing the existing one? The images that follow this aren't as wide. Is there some reason we need a wider image here?

Blackbaud-ToddRoberts: the intent was to use a somewhat realistic width for these full-page-context examples, but could shrink to e.g. 800px. Whether it fits next to the text or not is happenstance based on monitor size - on my screen the usage example is side by side.

blackbaud-johnly: Is the realistic width for full-page context a new thing? I don't see that pattern in the other docs we've updated. Phone field and tabs don't user larger full-page-context examples in the Usage section, and popover doesn't have any images in the Usage section. ... Anyway, I just thought this looked a little weird since all the other images work side-by-side for me except this one. And my browser is at full size on a fairly wide screen, so I think this will be the case for most users. No big deal if that's how we want to handle this, but I figured it was worth asking. Thanks.

Blackbaud-ToddRoberts: just pushed an update with narrower images

blackbaud-johnly: Sounds good.

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
...
+            <div class="caption caption-success sky-text-success">Use a split view when the user will take the same action for each item in the list.</div>
```
Blackbaud-AlexKingman: I'm seeing a narrow column of text above this first image. Is that deliberate?
Screen Shot 2019-07-17 at 2 17 19 PM

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
...
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+    </sky-demo-page-content>
+
+    <sky-demo-page-content sectionHeading="Behavior and states">
+      <h3>Automatically advance to the next item</h3>
+        <div class="sky-text-block">
+          <p>If the user is likely to work through the list of items sequentially, automatically advance to the next item in the list when the user invokes the primary action on the record.</p>
+        </div>
+      <h3>Set focus in the workspace</h3>
```
Blackbaud-AlexKingman: Is the spacing a bit off here?
Screen Shot 2019-07-17 at 2 23 12 PM

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
...
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <img src="~/assets/img/guidelines/split-view/behavior-resize.png" />
+          </sky-docs-visual-example-image>
+      </sky-docs-visual-example>
+    </sky-demo-page-content>
+    <h3>Responsiveness</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>In smaller viewports, the split view automatically switches from a side-by-side view to a panel-switching interaction where users switch between the list and the workspace. The responsive workspace header should include iterator buttons for users to move between items without going back to the list.</p>
```
blackbaud-johnly: I think "should include" should just be "includes," right? I didn't catch this the first time around, but this isn't a suggestion is it? This is just built-in functionality, I assume. If so, we should change "should include" to "includes"

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
...
+      Provides an observable to send commands to the split view component. The commands should respect <a stacheRouterLink="." fragment="skysplitviewmessage-type">the <sky-code>SkySplitViewMessage</sky-code> type</a>.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="SkySplitViewMessage type">
+    <sky-demo-page-property
+      propertyName="type"
+      isOptional="true"
+    >
+      Sets <sky-code>SkySplitViewMessage</sky-code> to one of the following types:
```
blackbaud-johnly: Are there plans to add more types?

src/app/components/split-view/index.html
blackbaud-johnly: This is the intended behavior. I was talking about it with Adam on Tuesday, and we agreed to leave this as is for now. I'll revisit this with him and Todd down the road.

src/app/components/split-view/index.html
blackbaud-johnly: More or less thing the same thing as with my last comment. This text fits the first column. Usually the image appears beside it, but in your browser, the image doesn't fit in that column. I think we should go with this as is for now and revisit it later.

src/app/components/split-view/index.html
Blackbaud-AlexKingman: Unsure at this point. We might have types for iterators down the road. Is there a better way to word this?

src/app/components/split-view/index.html
Blackbaud-AlexKingman: The iterators are no-longer built-in.

src/app/components/split-view/index.html
blackbaud-johnly: Given that there is only 1 item and we don't have current plans to add more, I wonder if it makes sense to change this to "Sets the type for `SkySplitViewMessage`. Currently, <sky-code>SkySplitViewMessageType.FocusWorkspace</sky-code>, which places focus on the first focusable element in the workspace, is the only available type." What do you think of that?

src/app/components/split-view/index.html
blackbaud-johnly: How do users include the iterators? I don't think I saw an input in the reference docs for this.

Blackbaud-AlexKingman: We originally had it as automatic portion of the workspace header but decided to pass on that functionality for MVP, as it adds quite a bit of complexity. Another approach we're discussing is to create an iterator component that users could include in their own templates. All that to say, is we haven't settled on a solution yet.

src/app/components/split-view/index.html
Blackbaud-AlexKingman: Yep that's fine with me!

src/app/components/split-view/index.html
blackbaud-johnly: In that case, should the docs include something like a recommendation for how to include the iterators? Or is that handled in the demo?

src/app/components/split-view/index.html
blackbaud-johnly: Done.

src/app/components/split-view/index.html
blackbaud-johnly: As per our discussion in standup, I removed the sentence about including iterator buttons since we don't have a method for including them. ... However, I noticed that the image includes iterator buttons. @Blackbaud-ToddRoberts, can you remove the iterator buttons from that screenshot?

Blackbaud-ToddRoberts: I thought it was possible for the consumer to add them even if we don't have an iterator button component yet? We want the UX docs to represent what should be there even if it requires manual intervention.

blackbaud-johnly: I thought from our discussion that it isn't possible to add iterator buttons, so we needed to update the docs to reflect that. However, if it is possible to add iterator buttons, my impression is that we'd need to update the docs to give some indication of how. @Blackbaud-AlexKingman, did I misunderstand? Can consumers add iterator buttons somehow?

Blackbaud-ToddRoberts: Example image is updated, sans iterator buttons

Blackbaud-AlexKingman:
Demo has been updated to use repeater's new "active" state. I think the only thing missing now is summary-action-bar upgrades.

Blackbaud-AlexKingman:
The demo has been updated to include `action-bar`.

Screen Shot 2019-08-15 at 3 20 44 PM

src/app/shared/docs-components/anatomy-item.component.html
```diff
@@ -1,7 +1,7 @@
 <ng-template #content><ng-content></ng-content></ng-template>
 
 <ng-container *ngIf="isOptional; else requiredStyle">
-  <div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">{{anatomyNumber}}</div><div class="anatomy-description"><ng-container *ngTemplateOutlet="content"></ng-container> <span class="sky-deemphasized">(optional)</span></div></div>
+  <div class="anatomy-item anatomy-item-optional"><div class="anatomy-marker">{{anatomyNumber}}</div><div class="anatomy-description"><ng-container *ngTemplateOutlet="content"></ng-container> <span class="sky-deemphasized"> (optional)</span></div></div>
```
Blackbaud-TrevorBurch: Just want to double check that we intended to add this space?

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
+<sky-tabbed-demo-page
+  pageTitle="Split view"
+>
+  <sky-demo-page-summary>
+    <p>
+      The split view component displays a list alongside a workspace where users can view details for the selected item and take actions.
+    </p>
+    <sky-alert alertType="warning">
+      This component is an initial release candidate.
```
Blackbaud-TrevorBurch: Is this warning too vague? Should we mention that, though not anticipated, changes to the component and its implementation detail may occur prior to official release? @Blackbaud-AlexKingman thoughts?

src/app/components/split-view/index.html
```diff
@@ -0,0 +1,239 @@
...
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="Split view child components">
+    <sky-demo-page-property
+      propertyName="sky-split-view-drawer"
+    >
+      Specifies the content to display in the split view's list panel.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="sky-split-view-workspace"
```
Blackbaud-TrevorBurch: We just added two child components of the workspace - the `sky-split-view-workspace-content` component and the `sky-split-view-workspace-footer` component. The footer component isn't always needed but consumers will always need to have the content component inside the main workspace component.

src/app/components/split-view/index.html
blackbaud-johnly: We can flesh out the statement if we want, but when Alex and I discussed this, we thought that Steve had been against calling out the possibility of changes since it's not supposed to change. But then on the Slack thread the other day, he seemed agreeable to fleshing this out. He also said to remove "initial."

So what do you guys think of fleshing this out to: "This component is a release candidate. We do not plan to change the component or its implementation details before the official release, but such changes are a possibility."

src/app/shared/docs-components/anatomy-item.component.html
blackbaud-johnly: This looks inadvertent to me, but I don't know. (If I made this change, it was definitely accidental. But I checked my commits, and I don't see this change there.)

src/app/components/split-view/index.html
blackbaud-johnly: I'll add entries for these components.

src/app/components/split-view/index.html
Blackbaud-ToddRoberts: Second sentence maybe "We do not plan to make any breaking changes to the component before the official release..."?

If we don't want to expand the text here, we should switch from an Alert to a Label saying "Release candidate"

blackbaud-johnly: I like the updated text, so I can make that change. But I'm good with a "Release candidate" label as well. Which do you prefer?

src/app/shared/docs-components/anatomy-item.component.html
blackbaud-johnly: I removed the extra space.

src/app/components/split-view/index.html
blackbaud-johnly: I updated the text. (Let me know if we should switch to a label.)

src/app/components/split-view/index.html
blackbaud-johnly: I added entries for the components.

Issue #279, opened 2018-01-10
Document guidelines for new page behavior


blackbaud-johnly:
Based on @Blackbaud-MattGregg's docs on navigation guidelines, create new topic in the Design section with guidance on the behavior of the back button. Paul said the behavior described in Matt's docs should be the default and that we just need documentation to call out that this is supposed to be the behavior unless something changes. Here's the notes form Matt's docs at https://blackbaudinc.app.box.com/s/9jmk38pmy2nb0zgrcoujyirwz55wcs17:

New page behaviors 
This section covers what interactions within SKY experiences should create a new URL and entry in the browser’s history to enable users to utilize the browser’s back button. 

Back button
Actions such as navigating to a completely new context (i.e., new record, list, section, etc.) should insert a new entry in the browser history with a new page title and enable a user to select the browser’s back button to go back to the previous page/view.

Sometimes this can be a judgement call. Use this rule of thumb:
If the majority of the page content changes or a user would perceive a significant change of context, create a new page in browser history.  

Inline interactions should not create a new URL or an entry in the browser’s history (e.g., steps in a wizard, opening detail panel, filtering/sorting)

Blackbaud-MattGregg:
Just a note this isn't accessibility-related. It's just how people expect the web sites and browser history/back buttons to function.

blackbaud-johnly:
Closing as outdated as we migrate docs tasks over to ADO. This was never a high priority, and when Denise looked into it, she ran into several issues determining what/how to document. Rather than starting over from scratch with something this old, I'm closing. We can always reopen if it turns out we really need some docs on this.

PR #568, opened 2019-07-15
[HOLD] Added content for back button guidelines

Blackbaud-DeniseCisneros:
Added content based on issue https://github.com/blackbaud/skyux2-docs/issues/279


src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">New page announcement for screenreaders</div>
+
+  <p>
+    Titles are announced as described in <a stacheRouterLink="." fragment="new-page-behavior">New page annoucement for screenreaders</a>.
+  </p>
+
+  <stache-page-anchor>
+    Tabs as containers for viewing items
```
blackbaud-johnly: Should this be "Tabs and vertical tabs"? Or does this section only apply to horizontal tabs?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </stache-page-anchor>
+
+  <p>
+    If tabs contain removable items such as lists or reports, don't include each tab in the browser history.
```
blackbaud-johnly: Same question on whether "tabs" at the start should be "tabs and vertical tabs." 

Also, I'd change the last part of the sentence to "... don't create entries in the browser history" to use the same terminology as we used above.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+	<div class="screenshot screenshot-info"> 
+    <img src="~/assets/img/guidelines/back-button/tabcontainers-backbutton.png" alt="Tabs as page navigation" />
```
blackbaud-johnly: I'm pretty sure we need to change the alt text here since this is not the section for tabs as navigation.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+	</div>
+
+<br>
+
+  <sky-row>
+    <sky-column>
+      <img class="screenshot screenshot-info" src="~/assets/img/guidelines/back-button/verticaltabs-backbutton.png" alt="Vertical tabs as navigation" />
```
blackbaud-johnly: I'm pretty sure we need to change the alt text here since this is not the section for vertical tabs as navigation.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+    </sky-column>
+  </sky-row>
+
+  <div class="sky-subsection-heading" style="margin-top: 20px;">Focus</div>
+
+  <p>
+    Based on guidance from <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel">WAI-ARIA Authoring Practices for tabs</a>, maintain focus on activated tabs.
```
blackbaud-johnly: I'd add "the" after "from." Also, should the end of the sentence be "tabs and vertical tabs" instead of just "tabs"?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">New page announcement for screenreaders</div>
+
+  <p>
+    This option doesn't require new page announcements. The selected tab receives an updated <sky-code>aria-selected="true"</sky-code> property which is communicated to assistive technologies.
```
blackbaud-johnly: I'd either add a comma before "which" or change it to "that."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <stache-page-anchor>
+    Modals (all sizes)
```
blackbaud-johnly: Do we need "(all sizes)"? Wouldn't that be the default assumption?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </stache-page-anchor>
+
+  <p>
+    When users open a modal from a SKY UX SPA, they should be able to use the back button to close the modal and return to the original context.
```
blackbaud-johnly: Do we need "from a SKY UX SPA" in the opening clause? I think we can just got with "When users open a modal, they should..."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">Avoid loss of work</div>
+
+  <p>
+    If a user selects the back button on a form with unsaved data, the SKY UX application asks the user if they want to close the form without saving their data or work using the confirmation dialog.
```
blackbaud-johnly: I'd change "a user selects" to "users select" and "the user" to "they." Also, we say "a form" and "the form," but I think this should be "a modal" and "the modal." ... I'd change "SKY UX application" to just "SPA." ... Also, as written, it sounds like the users are saving the work with the confirmation dialog. I'd move this up in the sentence: "If users select the back button on a modal with unsaved data, the SPA displays a dialog to confirm that they want to close the modal without saving their work."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">Focus</div>
+
+  <p>
+    In all circumstances, focus moves to an element contained in the dialog. Set the initial focus on the first focusable element unless a condition where doing otherwise is advisable.
```
blackbaud-johnly: We switch from "modal" to "dialog" here. I don't think we should do that. If I follow what we are saying here, I'd change the first sentence to "When a modal opens, always place focus on an element in the modal."

The where clause in the second sentence doesn't make sense. It reads like it's missing a word, and it's too vague. I'd change this sentence to "In general, place the initial focus on the first focusable element." unless there is some specific scenario we are referencing in the where clause.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <p>
+    If focus on the first interactive element causes the content to scroll out of view, add <sky-code>tabindex=-1</sky-code> to a static element at the top of a dialog. For example, add the code to a dialog title or the first paragraph to initially focus the element. For more best practices, see <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal">W3C's WAI-ARIA Authoring Practices for Dialog (Modal)</a>.
+  </p>
```
blackbaud-johnly: I think "focus" should either be "focusing" or "placing focus." Also, I think "a dialog" should be "the modal."

In the second sentence, I'd change "the code" to `tabindex=-1` and change "a dialog title" to "the title." I'd delete "the" before "first." and in the link, I'd change "Dialog (Modal)" to "modals." (In the tab links above, we just refer to them as "tabs" lowercase and generic, so I'd follow that pattern here or make "Tabs" uppercase above.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  <p>
+    This option doesn't require new page announcements.
+  </p>
+
+  <stache-page-anchor>
+    Maintaining context
+  </stache-page-anchor>
+
+  <p>
+    For inline interactions, determine the persistence level that makes sense for the change of state the user initiated. When users select the back button, return them to the context they left and don’t make them re-orient, and in some cases re-interact with the system, to return to where they were.
```
blackbaud-johnly: I'd change "the user initiated" to "that users initiate." ... Also, the second sentence is too long and it needs to be read a second time to read the stuff after "and." I'd make it a separate sentence: ".... left. Don’t make them re-orient, and in some cases re-interact with the system, to return to that context."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  </p>
+
+  <p>
+    If you expect users to maintain a specific context within a SKY UX app, retain the UI component states across sessions. Examples include:
```
blackbaud-johnly: I'd change "SKY UX app" to "SPA."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+    <li>
+      Tile expanded/collapsed states
+    </li>
+    <li>
+      Tab selected (list or tabs)
+    </li>
+    <li>
+      Lists
+    </li>
+    <ul>
```
blackbaud-johnly: The spacing here is weird. The second list has more space between the items. I think the problem is that the `ul` tag and all of its contents should be within the Lists `li` tag. I'd try putting it there to see if that fixes the spacing.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+    For inline interactions, determine the persistence level that makes sense for the change of state the user initiated. When users select the back button, return them to the context they left and don’t make them re-orient, and in some cases re-interact with the system, to return to where they were.
+  </p>
+
+  <p>
+    If you expect users to maintain a specific context within a SKY UX app, retain the UI component states across sessions. Examples include:
+  </p>
+
+  <ul>
+    <li>
+      Tile position
```
blackbaud-johnly: Aren't these first 2 items really sub-items for tiles? We have "Lists" and then a series of list sub-items. Should we have "Tiles" and two sub-items as well? Or should we get rid of the "Lists" parent bullet and make everything on the same level? After all, most of the list sub-items still mention "list"

Also, seems like this should be "Positions of tiles" since I don't think we're talking about a specific UI component but all UI components.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+        Sort order for a specific list
+      </li>
+      <li>
+        Column choices for a specific list
+      </li>
+      <li>
+        View state for a specific list
+      </li>
+      <li>
+        Grid view, compact and regular, across lists
```
blackbaud-johnly: I don't follow what this means.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+      </li>
+    </ul>
+    <li>
+      Page context (e.g. user, credit card, environment)
```
blackbaud-johnly: I'd make "contexts" plural and remove the parentheses and change "e.g." to "such as." ... Also, credit card is a page context?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+  <p>
+    If you expect users to maintain a specific context within a SKY UX app, retain the UI component states across sessions. Examples include:
+  </p>
+
+  <ul>
+    <li>
+      Tile position
+    </li>
+    <li>
+      Tile expanded/collapsed states
```
blackbaud-johnly: Seems like this would be more clear as "Expanded and collapsed states of tiles"

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+    </li>
+    <li>
+      Tab selected (list or tabs)
```
blackbaud-johnly: Seems like this would be more clear as "Selected tabs" ... I don't follow what the parenthetical means. What does the selected tab have to do with lists?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+    </li>
+    <li>
+      Lists
+    </li>
+    <ul>
+      <li>
+        Sort order for a specific list
```
blackbaud-johnly: Seems like this should be "Sort order for lists"

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+      </li>
+      <li>
+        Column choices for a specific list
```
blackbaud-johnly: What are "column choices"? Also, I think "for a specific list" should be "for lists."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
...
+      </li>
+      <li>
+        View state for a specific list
```
blackbaud-johnly: Seems like this should be "View state for lists"

src/app/design/guidelines/back-button/index.html
Blackbaud-DeniseCisneros: I'm going to leave this as tabs because the section it links to only uses tabs. I don't _think_ that'd cause confusion for anyone but just in case they think they've landed on the wrong section.

src/app/design/guidelines/back-button/index.html
Blackbaud-DeniseCisneros: I assume that's the default assumption too so I made the change 👍

src/app/design/guidelines/back-button/index.html
Blackbaud-DeniseCisneros: I removed "compact and regular," but it might need some more additional clarification if it's still unclear

src/app/design/guidelines/back-button/index.html
Blackbaud-DeniseCisneros: Not sure what "column choices" mean, but I did edit it to "Column choices for lists"

Blackbaud-DeniseCisneros:
The items I didn't resolve are items I wasn't fully sure of and could use @Blackbaud-MattGregg's input.

src/app/design/guidelines/back-button/index.html
blackbaud-johnly: Sounds good. Thanks.

src/app/design/guidelines/back-button/index.html
blackbaud-johnly: Sounds good. Thanks.

src/app/design/guidelines/back-button/index.html
blackbaud-johnly: I assume we are referring to settings such as column widths, sort order, etc. I wonder if "settings" instead of "choices" would work better here. @Blackbaud-MattGregg, can you offer some clarification here?

src/app/design/guidelines/back-button/index.html
blackbaud-johnly: I'm still not entirely sure what this means. I've searched the grid and list view grid docs, and I can't tell what we are talking about here. If we don't know what this means, I'd delete this bullet. @Blackbaud-MattGregg, can you provide some clarification here?

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,165 @@
+<stache
```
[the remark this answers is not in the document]
blackbaud-johnly: @Blackbaud-MattGregg and @Blackbaud-ToddRoberts, what do you guys think about the location of these docs. I'm not advocating moving this to the Accessibility section, but I want to make sure we have it in the right place. Thanks.

Blackbaud-MattGregg:
@blackbaud-johnly Curious if something sparked the creation of these guidelines?
Given our content strategy of centering design guidelines mainly around the components. I think much of what's here should go on the tabs and modal component pages.

Blackbaud-MattGregg:
@blackbaud-johnly @Blackbaud-ToddRoberts For the sections on Page announcements for screen reader users and Maintaining state, those were items in the proposal that needed to be built into the framework I believe. Have these been added to the system so that teams get them out of the box are can implement them themselves?

For the new page behaviors, I'm not sure the best spot for this. It's speaking a bit more to the technical implementation which could be developer content but also good for designers/others to know this aspect of the overall web interaction model.

blackbaud-johnly:
@Blackbaud-MattGregg, these docs were created to address https://github.com/blackbaud/skyux2-docs/issues/279, which I created last year after meeting with you and Todd. I never got to it, but Denise picked it up on our backlog.

Blackbaud-ToddRoberts:
The tab & modal specific content should live on those component pages. Not sure where the New Page stuff should go; design > guidelines doesn't feel right since the content is more relevant to developers. Is there an equivalent developer section that has this sort of "baseline" documentation that is not component-specific?

Blackbaud-DeniseCisneros:
I moved the component specific content to their component pages and tried to place it where it made sense. Styling for those items varies a little, so I'd love some feedback on that. 

@Blackbaud-ToddRoberts to your question, I had trouble finding a generic enough section for the New page behavior and Maintaining context content. I was thinking maybe Guidelines > Page design for the former, and shortening the latter to fit in the UI Accessibility Checklist?

src/app/components/modal/index.html
```diff
@@ -150,7 +150,9 @@
   <sky-demo-page-content sectionHeading="Accessibility">
     <p>Follow these guidelines to ensure that people can use modals with assistive technology or keyboards.</p>
     <ul>
-      <li>When a modal opens, focus should be on the first focusable item in the modal. If the modal does not include focusable items, then focus should be on the modal container element.</li>
+      <li>When a modal opens, focus should be on the first focusable item in the modal. If the modal does not include focusable items, then focus should be on the modal container element.
```
blackbaud-johnly: This line should have a closing `li` tag

src/app/components/modal/index.html
```diff
@@ -150,7 +150,9 @@
...
+      <li>If placing focus on the first interactive element causes the content to scroll out of view, add <sky-code>tabindex=-1</sky-code> to a static element at the top of a modal. For example, add <sky-code>tabindex=-1</sky-code> to the title or first paragraph to initially focus the element. For more best practices, see <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal">W3C's WAI-ARIA Authoring Practices for modals</a>.
```
blackbaud-johnly: Should we have "on" after "initially focus" in the second sentence? Seems like to me, but I'm not sure.

Also, is the link in the last sentence really specific to this bullet? It doesn't seem like it to me, so I'd suggest moving that last line to make it a standalone paragraph after the bullet list.

src/app/components/modal/index.html
```diff
@@ -194,6 +196,9 @@
 				<li>Use a primary <a routerLink="/components/button">button</a> for the most important action the user will take. Use secondary buttons for less important actions. Use a link button for the Cancel action.</li>
 				<li>If a modal has only one action, you should generally use the <a routerLink="/components/confirm">confirm</a> component instead.</li>
 			</ul>
+
+    <div class="sky-subsection-heading">Back button</div>
+      <p>When users open a modal, they should be able to use the back button to close the modal and return to the original context. If users select the back button on a modal with unsaved data, the SPA displays a dialog to confirm that they want to close the modal without saving their work.</p>
```
blackbaud-johnly: It's minor, but the opening clause seems unnecessary. I'd start this sentence as "Users should be able to..."

Also, I think we can tighten up `If users select the back button on a modal with unsaved data` in the second sentence to `If the modal includes unsaved data`.

Finally, the verb in the first sentence is conditional ("should be able to") while the verb in the second sentence is not ("displays"). It seems like both sentences should either be conditional or not. I mean, "should be able to" suggests that maybe sometimes the back button doesn't work this way. If so, then the second sentence should be conditional to, i.e. "should display." But if the first statement is not actually conditional, then I'd suggest changing "should be able to" to "can."

src/app/components/tabs/index.html
```diff
@@ -142,6 +142,16 @@ <h3>Page titles</h3>
 			<p>
 				For information about how to use <a href="https://angular.io/docs/ts/latest/cookbook/set-document-title.html">the Angular <sky-code>Title</sky-code> service</a> and the SKY UX <sky-code>title</sky-code> configuration property to specify the title, see <a stacheRouterLink="learn/reference/configuration" fragment="app">the configuration reference</a>.
 			</p>
+
+			<h3>Tabs as site-level navigation</h3>
```
blackbaud-johnly: Is this section supposed to be a child of the Page titles section? If so, I think this should be an h4 sine Page titles is an h3.

And speaking of the header, is there a reason it's not part of the includes file? That seems a little weird to me. But maybe it makes sense if we expect the header to be different sizes in different places.

src/app/components/tabs/index.html
```diff
@@ -142,6 +142,16 @@ <h3>Page titles</h3>
...
+
+			<stache-include fileName="/components/tabs-sitelevel.html">
+			</stache-include>
+
+			<h3>Tabs as containers for viewing items</h3>
```
blackbaud-johnly: Is this section supposed to be a child of the Page titles section? If so, I think this should be an h4 sine Page titles is an h3.

Also, I'd change "for viewing" to "to view" to avoid the gerund. ... And speaking of the header, is there a reason it's not part of the includes file? That seems a little weird to me.

src/stache/includes/components/tabs-sitelevel.html
```diff
@@ -0,0 +1,19 @@
+<p>
+  If you use tabs or vertical tabs as page navigation, each one should have an entry in the browser history with a distinct title. This enables users to select the browser's back button to return to the previous view. Tabs and vertical tabs that are used as site-level navigation contain all of a page's content and can't be dismissed.
```
blackbaud-johnly: I'd change "each one" to "each tab." ... But I wonder if there is a way to reword the portion of the first sentence after the comma to make it a command and avoid "should," which could make it sound optional. Maybe "provide an entry for each tab in the browser history with a distinct title"? What do you think?

src/stache/includes/components/tabs-sitelevel.html
```diff
@@ -0,0 +1,19 @@
...
+</p>
+
+<div class="screenshot screenshot-info">
+  <img src="~/assets/img/guidelines/back-button/tabs-backbutton.png" alt="Tabs as page navigation" style="width: 100%" />
+</div>
+
+<div class="sky-subsection-heading" style="margin-top: 20px;">Focus</div>
+
+<p>
+  Based on guidance from the <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel">W3C's WAI-ARIA Authoring Practices for tabs</a>, maintain focus on activated tabs.
```
blackbaud-johnly: I'd flip the order of this sentence to get the recommendation up front and the link at the end: "Maintain focus on activated tabs as recommended in the <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel">W3C's WAI-ARIA Authoring Practices for tabs</a>."

src/stache/includes/components/tabs-sitelevel.html
```diff
@@ -0,0 +1,19 @@
...
+</p>
+
+<div class="sky-subsection-heading">Screen readers</div>
+
+<p>
+  Titles are announced as described in <a stacheRouterLink="." fragment="new-page-behavior">New page behavior</a>.
```
blackbaud-johnly: What is this link supposed to be pointing to? I don't think we can use a link like this in an includes item because not every page is going to have a New page behavior section. The tabs topic doesn't, so this link just point to the top of the page in that topic.

src/stache/includes/components/tabs-containers.html
```diff
@@ -0,0 +1,27 @@
+<p>
+  If tabs and vertical tabs contain removable items such as lists or reports, don't create entries in the browser history.
+</p>
+
+<div class="screenshot screenshot-info"> 
+  <img src="~/assets/img/guidelines/back-button/tabcontainers-backbutton.png" alt="Tabs as containers for viewing items" />
```
blackbaud-johnly: We have 2 images here, while we only have 1 image (no vertical tabs image) in the previous section. Is this correct?

src/stache/includes/components/tabs-containers.html
```diff
@@ -0,0 +1,27 @@
...
+<sky-row>
+  <sky-column>
+    <img class="screenshot screenshot-info" src="~/assets/img/guidelines/back-button/verticaltabs-backbutton.png" alt="Vertical tabs as containers for viewing items" />
+  </sky-column>
+</sky-row>
+
+<div class="sky-subsection-heading" style="margin-top: 20px;">Focus</div>
+
+<p>
+  Based on guidance from the <a href="https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel">WAI-ARIA Authoring Practices for tabs</a>, maintain focus on activated tabs.
```
blackbaud-johnly: I'd flip the order of this sentence to get the recommendation up front and the link at the end: "Maintain focus on activated tabs as recommended in the W3C's WAI-ARIA Authoring Practices for tabs."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+  pageTitle="Back Button"
+  navTitle="Back button"
+  showTableOfContents="true">
+
+  <stache-page-anchor>
+    New page behavior
+  </stache-page-anchor>
+
+  <p>
+    When users navigate to a new context such as a new record, list or section, the action should create an entry in the browser history with a new page title. This entry enables users to select the browser's back button to return to the previous page or view.
```
blackbaud-johnly: I don't think we need the second "new" before "record."

I might be wrong, but I think "page or view" at the end should just be "view."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+  </p>
+  <p>
+    As a rule of thumb, create a new page in the browser history if the majority of the page content changes or users perceive a signficant change of context. Don't create a new URL or new browser entry for an inline interaction such as steps in a wizard or opening a detail panel.
```
blackbaud-johnly: should "a new page" be "an entry"?

Also, I don't think we need "opening" near the end.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">Focus</div>
+
+  <p>
+    After loading a new page, users should be able to use the tab key to reach the first focusable element on the page.
```
blackbaud-johnly: I'd delete "new" in the opening clause.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+  </p>
+
+  <div class="sky-subsection-heading">Screen readers</div>
+
+  <p>
+    When a page loads, it should behave similarly to what users expect from a regular server-side application. The page should announce the page's <sky-code>&lt;title&gt;</sky-code> to the screen reader and inform users that it loaded via an ARIA live region when the <sky-code>&lt;title&gt;</sky-code> element is ready. Behaviors vary based on different browser and screen reader combinations, but SKY UX applications use this combination.
+  </p>
+
+
+  <div class="sky-subsection-heading">Focus</div>
```
blackbaud-johnly: Delete this extra header.

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+
+  <stache-page-anchor>
+    Maintaining context
+  </stache-page-anchor>
+
+  <p>
+    For inline interactions, determine the persistence level that makes sense for the change of state that user initiates. When users select the back button, return them to the context they left. Don't make them re-orient, and in some cases re-interact with the system, to return to that context.
```
blackbaud-johnly: I'd change "user initiates" to "users initiate"

I wonder if we can tighten up the last sentence to "Don't make them re-orient or re-interact with the system to return to that context."

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+      Column choices for lists
+    </li>
+    <li>
+      View state for lists
+    </li>
+    <li>
+      Grid view across lists
+    </li>
+    <li>
+      Page contexts such as user, credit card, environment
```
blackbaud-johnly: Seems like we should have "and" or "or" after the second comma

src/app/design/guidelines/back-button/index.html
```diff
@@ -0,0 +1,71 @@
...
+  <div class="sky-subsection-heading">Focus</div>
+
+  <p>
+    After loading a new page, users should be able to use the tab key to reach the first focusable element on the page.
+  </p>
+
+  <div class="sky-subsection-heading">Screen readers</div>
+
+  <p>
+    When a page loads, it should behave similarly to what users expect from a regular server-side application. The page should announce the page's <sky-code>&lt;title&gt;</sky-code> to the screen reader and inform users that it loaded via an ARIA live region when the <sky-code>&lt;title&gt;</sky-code> element is ready. Behaviors vary based on different browser and screen reader combinations, but SKY UX applications use this combination.
```
Blackbaud-MattGregg: @blackbaud-johnly @Blackbaud-ToddRoberts Has this functionality been added to SKYUX? This proposal was to add a common solution to the framework so team's wouldn't have to do this one off. If it hasn't been added, then I would wait to include this. Also, is there an issue for it in the backlog?

Blackbaud-MattGregg: Actually just found there is an issue and this isn't built yet. I think we should remove this until we offer a way for the system to actualize it. https://github.com/blackbaud/skyux-core/issues/67

Blackbaud-DeniseCisneros:
Closing because this PR is severely out of date but keeping the issue open in case we need to revisit it.

PR #587, opened 2019-07-24
Toggle switch

Blackbaud-SteveBrush:
Issue: https://github.com/blackbaud/skyux2-docs/issues/586


src/app/components/toggle-switch/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+  <sky-demo-page-properties
+    sectionHeading="Toggle switch properties"
+  >
+    <sky-demo-page-property
+      propertyName="ariaLabel"
+      defaultValue="undefined"
+      [isOptional]="true"
+    >
+      Defines a string value to label the toggle switch and set the <sky-code>aria-label</sky-code> attribute to support <a stacheRouterLink="." fragment="accessibility">accessibility</a>. This property accepts <sky-code>string</sky-code> values.
```
blackbaud-johnly: We link to an accessibility section that doesn't exist. Is that still coming?

Also, we set the default value to `undefined`. For other `ariaLabel` properties, we don't list a default value. Do we need to add `undefined` as the default value for those properties or remove it here?

Also, I know the wording in that first sentence corresponds to what we already have in place for other `ariaLabel` properties, but I wonder if we could tighten it up here and in those other components. The last sentence explains that this accepts string values, so would we lose anything by changing "Defines a string value to label the toggle switch..." to just "Labels the toggle switch..."? ... If we make that change here, we'd also need to change "set" to "sets." That said, in the other `ariaLabel` properties, I don't see explicit references to the `aria-label` attribute in the property descriptions. Should those properties also mention this? Or should we remove it here for consistency?

Finally, should this property make some reference to the `sky-toggle-switch-label` component? I assume so. Other components have `ariaLabel` and `ariaLabelledBy` properties that sorta work in conjunction. Since we don't have `ariaLabelledBy` here, I assume `sky-toggle-switch-label` is the default element to label the toggle switch and that you wouldn't need to label it and also use `ariaLabel`. But I'm not sure. it seems like we should mention `sky-toggle-switch-label` somewhere in the docs, so maybe this is a logical place.

src/app/components/toggle-switch/index.html
```diff
@@ -0,0 +1,77 @@
...
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties
+    sectionHeading="Toggle switch events"
+  >
+    <sky-demo-page-property
+      propertyName="toggleChange"
+      [isOptional]="true"
+    >
+      Fires when the checked state of a toggle switch changes. The event provides an argument of type <sky-code>SkyToggleSwitchChange</sky-code>.
```
blackbaud-johnly: When types like this have a separate section, I've been linking to them. I'd reorder the second sentence to "The event provides an argument with the `SkyToggleSwitchChange` type" and make "the `SkyToggleSwitchChange` type" a link to the related section. (I know the section is nearby, but I'd add the link just for consistency.)

src/app/components/toggle-switch/index.html
```diff
@@ -0,0 +1,77 @@
...
+    >
+      References the <sky-code>checked</sky-code> state of the toggle switch.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-example>
+    <sky-toggle-switch-demo>
+    </sky-toggle-switch-demo>
+    <sky-demo-page-code
+      demoName="Toggle switch"
```
blackbaud-johnly: The demo seems a little off to me. See my notes below about the first two toggle switches.

src/app/demos/toggle-switch/toggle-switch-demo.component.html
```diff
@@ -0,0 +1,60 @@
+<p>
+  <sky-toggle-switch
+    [checked]="true"
```
blackbaud-johnly: The first two toggle switches seem weird to me. When I click the first toggle switch to switch it from "on" to "off," the text beside it remains 'Switch is "on" ' Likewise, when I click the second toggle switch to switch it from "off" to "on," the text beside it remains 'Switch is "off" ' So basically, as soon as someone interacts with those toggle switches, the labels are inaccurate. I think it would sense to combine these toggle switch into a single toggle switch with a label that alternates between 'Switch is "on" ' and 'Switch is "off" ' But if that's not feasible, I'd change the existing first two toggle switches to make them non-interactive so that we don't have the incorrect labels as soon as someone selects them.

src/app/demos/toggle-switch/toggle-switch-demo.component.html
```diff
@@ -0,0 +1,60 @@
...
+  </sky-toggle-switch>
+</p>
+
+<p>
+  <sky-toggle-switch
+    [checked]="true"
+    [disabled]="true"
+  >
+    <sky-toggle-switch-label>
+      Switch is "on", readonly
```
blackbaud-johnly: "read only" is 2 words. And I'd change the comma to "and" here

src/app/demos/toggle-switch/toggle-switch-demo.component.html
```diff
@@ -0,0 +1,60 @@
...
+  </sky-toggle-switch>
+</p>
+
+<p>
+  <sky-toggle-switch
+    [checked]="false"
+    [disabled]="true"
+  >
+    <sky-toggle-switch-label>
+      Switch is "off", readonly
```
blackbaud-johnly: "read only" is 2 words. And I'd change the comma to "and" here

src/app/components/toggle-switch/index.html
Blackbaud-DeniseCisneros: Should the accessibility link be to this section?

blackbaud-johnly: Usually when we have links like this in `ariaLabel` properties, it points to an accessibility section within the same topic, like in the radio button topic at https://developer.blackbaud.com/skyux/components/radio.

Blackbaud-ToddRoberts:
is this superceded by #591?

src/app/components/toggle-switch/index.html
```diff
@@ -0,0 +1,200 @@
...
+      </sky-demo-page-module-info>
+
+      <sky-demo-page-properties
+        sectionHeading="Toggle switch properties"
+      >
+        <sky-demo-page-property
+          propertyName="ariaLabel"
+          [isOptional]="true"
+        >
+          Sets the <sky-code>aria-label</sky-code> attribute of the toggle switch. Only use this value if the `sky-toggle-switch-label` component is not used. This property accepts <sky-code>string</sky-code> values.
```
Blackbaud-DeniseCisneros: Should sky-toggle-switch-label be between the <sky-code> tag, or are the tick marks intentional?

Blackbaud-SteveBrush: Oops! Good catch.

Issue #595, opened 2019-07-26
Document `showAddNewRecordButton` input for `SkySelectFieldComponent`


Blackbaud-SteveBrush:
Pull request for the feature: https://github.com/blackbaud/skyux-select-field/pull/30

Blackbaud-DeniseCisneros:
Merged and closed in https://github.com/blackbaud/skyux2-docs/pull/599

Issue #592, opened 2019-07-26
Document the Wait Service `nonBlockingWrap` and `blockingWrap` methods


Blackbaud-SteveBrush:
Pull request: https://github.com/blackbaud/skyux-indicators/pull/39/files

Blackbaud-DeniseCisneros:
Merged and closed in https://github.com/blackbaud/skyux2-docs/pull/599

PR #599, opened 2019-08-05
Updates to wait and select field components

Blackbaud-DeniseCisneros:
Added new method to the wait component and the new showAddNewRecordButton property to the select field 

Addresses https://github.com/blackbaud/skyux2-docs/issues/592 and https://github.com/blackbaud/skyux2-docs/issues/595


src/app/components/select-field/index.html
```diff
@@ -63,6 +63,13 @@
     >
     Specifies the selection mode that determines whether users can select one item or multiple items in the select field component's modal. The valid options are <sky-code>single</sky-code>, which displays a text input and allows users to select one item, and <sky-code>multiple</sky-code>, which displays a button and allows users to select multiple items.
     </sky-demo-page-property>
+    <sky-demo-page-property
```
blackbaud-johnly: We also need to document the new `addNewRecordButtonClick` event. In the PR where the input was added, the output is at https://github.com/blackbaud/skyux-select-field/pull/30/files#diff-d6e48fc07613709745cf9d2b62b7fa31R120.

src/app/components/select-field/index.html
```diff
@@ -63,6 +63,13 @@
...
+      propertyName="showAddNewRecordButton"
+      isOptional="true"
+      defaultValue="false"
+    >
+    Indicates whether to display a button with a <sky-code>New</sky-code> text label. Users can add new items directly from the select field modal. This property accepts <sky-code>boolean</sky-code> values.
```
blackbaud-johnly: I don't think the code snippet style for the text label follows our general style. I think we generally just refer to buttons generically with no formatting. Also, what do you think of combining the first two sentences to tighten up? For example: "Indicates whether to display a button for users to add new items directly from the select field modal."

src/app/components/wait/index.html
```diff
@@ -51,6 +51,12 @@
     >
```
blackbaud-johnly: Should we also update the wait service methods to include the public `dispose` method at https://github.com/blackbaud/skyux-indicators/blob/master/src/app/public/modules/wait/wait.service.ts#L64? I assume this is a question for @Blackbaud-SteveBrush.

src/app/components/wait/index.html
```diff
@@ -51,6 +51,12 @@
...
     Starts a non-blocking page wait on the page.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="blockingWrap"
+      isOptional="true"
+    >
+    Launches a page wait and automatically stops when the specific asynchronous event completes. The most common use case includes backend calls.
```
blackbaud-johnly: For consistency with the rest of the descriptions, I think we need "blocking" at the start: "Launches a blocking page wait..."

Also, is the last sentence correct? Just making sure that the most common use case for blocking page waits is the same as for non-blocking page waits since the last sentence of that method is identical to the one here.

src/app/components/wait/index.html
Blackbaud-SteveBrush: Good catch, but I think we're okay to not include it in the docs. It's primarily used internally.

src/app/components/wait/index.html
Blackbaud-DeniseCisneros: I believe Colby contributed this so I checked in with him when adding the properties. All that to say, the second line doesn't really match existing docs so I'm okay to leave out the second sentence.

blackbaud-johnly: Well, if the most common use case is the same for both and it is useful to point it out, I think it's fine to keep that sentence. I just wasn't sure if the use case would be the same for both. Thanks!

src/app/components/wait/index.html
blackbaud-johnly: OK. Thanks!

src/app/components/select-field/index.html
```diff
@@ -88,6 +95,12 @@
   </sky-demo-page-properties>
 
   <sky-demo-page-properties sectionHeading="Select field events">
+    <sky-demo-page-property
+      propertyName="addNewRecordButtonClick"
+      isOptional="true"
+    >
+      Fires when users add new items directly from the select field modal.
```
Blackbaud-SteveBrush: This event doesn't fire when new items are created (the consumer must implement that logic themselves), it only fires when the New button is clicked. Is it worth pointing out that distinction?

src/app/components/select-field/index.html
```diff
@@ -63,6 +63,13 @@
     >
     Specifies the selection mode that determines whether users can select one item or multiple items in the select field component's modal. The valid options are <sky-code>single</sky-code>, which displays a text input and allows users to select one item, and <sky-code>multiple</sky-code>, which displays a button and allows users to select multiple items.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="showAddNewRecordButton"
+      isOptional="true"
+      defaultValue="false"
+    >
+    Indicates whether to display a button for users to add new items directly from the select field modal.
```
Blackbaud-SteveBrush: This will only show the New button; it doesn't automatically add items to the list (the consumer must provide that logic themselves). Is it worth pointing out that distinction?

blackbaud-johnly: It seems worth pointing out the distinction to me. What do you think of adding a second sentence like: "Indicates whether to display a button for users to add new items directly from the select field modal. You must provide the logic for adding items."?

src/app/components/select-field/index.html
Blackbaud-DeniseCisneros: I think it is. That was my best guess and was hoping to be corrected or validated during review. 😆

PR #610, opened 2019-08-13
Add container colors

Blackbaud-AdamFunderburk:
Adding 7 new container colors to the Design>Styles>Colors page

Dependent on this PR in skyux-design-tokens
https://github.com/blackbaud/skyux-design-tokens/pull/26


src/app/design/styles/color/index.html
```diff
@@ -355,6 +355,156 @@
 	</div>
 		
 	<div>
+  <stache-page-anchor>
+    Containers
+  </stache-page-anchor>
+    <p>
+    Use container colors when categorizing blocks of content that will be mixed together, such as a calendar with different types of events. Container colors can be chosen specifically by using their variable name, or applied in a sequence.
```
blackbaud-johnly: It's minor, but I'd tighten up "when categorizing blocks of content that will be mixed together" to "to categorize blocks of content that mix together." ... I'd also change to "such" as clause to make it a separate sentence: "For example, use container colors for different types of events in calendars."

In the last sentence, we don't need the comma. Also, it's minor, but I'd tighten up "specifically by using their variable name, " to "by variable name"

src/app/design/styles/color/index.html
```diff
@@ -355,6 +355,156 @@
...
+    </p>
+        
+  <div class="definition-table definition-table-wide">
+    <div class="definition-item">
+      <div class="swatch" style="background-color: #bfecfb"></div>
+    </div>
+    <div class="definition-props">
+      <div class="definition-usage">
+        <div class="sky-field-label">Usage</div>
+        <div>Background color for categorized containers or for containers that have no categories.</div>
```
blackbaud-johnly: For consistency with the existing descriptions on the page, delete "color." ... Same suggestion for the rest of the entries in this section.

This description confuses me. As written, it sounds like this is the the background for containers that include categories or for containers without categories, which sounds like this is the background for all containers, right? But I assume we are making a distinction between things that just isn't clear. At first I thought the problem was "categorized containers" and that it should be "categories within containers"? I **think** that might be what we mean, but I'm not sure. I think we are using "containers" to mean two different things here. The intro description makes it sound like containers ARE the categories, but this description makes it sound like containers contain the categories. So we seem to be using "containers" to mean both the categories within a container and the container itself. Either that or I'm just missing something. ... Either way, I have the same question/suggestion for the rest of the entries in this section.

Blackbaud-AdamFunderburk: Maybe we get more clarity by labeling this set of colors "Categorizations" instead of "Containers". The explanation (and description of each row) probably gets simpler if we just start with that framing.

src/app/design/styles/color/index.html
```diff
@@ -356,10 +356,10 @@
 		
 	<div>
   <stache-page-anchor>
-    Containers
+    Categorization
   </stache-page-anchor>
     <p>
-    Use container colors to categorize blocks of content that mix together. For example, use container colors for different types of events in calendars. Container colors can be chosen by variable name or applied in a sequence.
+    Use categorization colors to categorize blocks of content that mix together. For example, use categorization colors for different types of events in calendars. These colors can be chosen by variable name or applied in a sequence.
```
blackbaud-johnly: I'd change "that mix together" to "within the same container"

And I'd tweak the last sentence to use active voice: "You can select colors for content categories by variable name or apply them in a sequence."

src/app/design/styles/color/index.html
```diff
@@ -369,7 +369,7 @@
     <div class="definition-props">
       <div class="definition-usage">
         <div class="sky-field-label">Usage</div>
-        <div>Background for categorized containers or for containers that have no categories.</div>
+        <div>Background for categorized content.</div>
```
blackbaud-johnly: Just curious, did we decide that this is NOT the background color for empty containers?

src/app/design/styles/color/index.html
```diff
@@ -369,7 +369,7 @@
...
       </div>
       <div class="definition-code">
         <div class="sky-field-label">Code</div>
```
blackbaud-johnly: Is it too late to rename the styles for the colors to use "categorization" or "category" instead of "container"?

Blackbaud-AdamFunderburk: Nope. Done.

src/app/design/styles/color/index.html
blackbaud-johnly: Is removing the "no categories" part from the original version correct?

src/app/design/styles/color/index.html
```diff
@@ -487,6 +489,156 @@
   </div>
 	
 	</div>
+    
+    <div>
+  <stache-page-anchor>
+    Categorization
+  </stache-page-anchor>
+    <p>
+    Use categorization colors to categorize blocks of content by significant type. For example, use categorization colors for different types of events in calendars.
```
blackbaud-johnly: It might just be me (and it might be too nitpicky to worry about), but the wording of `by significant type` sounds off to me, and it sees a little vague. What is a significant type? What do you think about changing `by significant type` to `based on shared types`?

src/app/design/styles/color/index.html
Blackbaud-AdamFunderburk: I believe it is. I think the scenario where it applies is too specific and edge-case-y to worry about in the description at this atomic level.

src/app/design/styles/color/index.html
Blackbaud-AdamFunderburk: I think "significant" is more informative here. It's true that there's no reusable metric for "significance", but using that word at least _aims_ for the point that the distinction between the categories should be an important one for the user.

src/app/design/styles/color/index.html
blackbaud-johnly: Sounds good. Thanks. Just wanted to make sure.

src/app/design/styles/color/index.html
Blackbaud-AdamFunderburk: Although, "based on" may be the clearer preposition here.

blackbaud-johnly: Works for me. Thanks!

blackbaud-johnly: Took another look at this because the tab was still open and mulled this over again. No sweat if you want to leave this as is, but what do you think of changing `by significant type` to `when the distinction between types of content is significant to users`? I think that might make this a little clearer.

Blackbaud-AdamFunderburk: Done.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-add-container-colors](https://developer.blackbaud.com/skyux?_sp.spa=pr-add-container-colors)

Issue #590, opened 2019-07-25
Document new input for repeater component


blackbaud-johnly:
For https://github.com/blackbaud/skyux-lists/pull/51, update the repeater docs to include entries for the `activeIndex` property. 

Note that in the changes, it looks like the `inlineFormConfig` and `inlineFormTemplate` inputs and the `inlineFormClose` output may also have changed.

Blackbaud-AlexKingman:
@blackbaud-johnly 

> Note that in the changes, it looks like the inlineFormConfig and inlineFormTemplate inputs and the inlineFormClose output may also have changed.

No docs changes needed for these properties. Those changes were simply reordering to be in compliance with our code standards. These properties still work the same as before.

Blackbaud-AlexKingman:
Suggestion for `activeIndex`:

Specifies the index of the active repeater item. Only one item can be active at a time. This property accepts `number` values. Optional. Default: `undefined`

Blackbaud-DeniseCisneros:
Fixed and merged in https://github.com/blackbaud/skyux2-docs/pull/612

PR #612, opened 2019-08-14
Added activeIndex property

Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/590

### DIFF
--- a/src/app/components/repeater/index.html
+++ b/src/app/components/repeater/index.html
@@ -17,6 +17,7 @@
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="expandMode"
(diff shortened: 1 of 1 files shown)


src/app/components/repeater/index.html
```diff
@@ -26,6 +26,11 @@
       </li>
     </ul>
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="activeIndex"
```
blackbaud-johnly: This should be the first item in the table because it comes first alphabetically. ... I'll pull down the branch and fix this, but adding a comment in case you are still working on this.

Issue #606, opened 2019-08-09
Document router commands for permalink property in action button component


blackbaud-johnly:
Update the docs at https://developer.blackbaud.com/skyux/components/action-button#skyactionbuttonpermalink-properties to address an internal question about the available options for the "commands property to specify router commands." ... Filing this issue based on internal feedback. The internal-only Slack thread where the question was raised is at https://app.slack.com/client/T02BBN218/C3TQ4C0KG/thread/C3TQ4C0KG-1565364685.448900.

Blackbaud-DeniseCisneros:
Fixed and merged in https://github.com/blackbaud/skyux2-docs/pull/620

PR #620, opened 2019-08-22
Added info about commands property

Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/606

I'm not in love with the way I handled the addition; definitely open to feedback.

### DIFF
--- a/src/app/components/action-button/index.html
+++ b/src/app/components/action-button/index.html
@@ -72,7 +72,7 @@
       propertyName="route"
       isOptional="true"
     >
-      Specifies an Angular router link for the action button. This property accepts a custom object with a <sky-code>commands</sky-code> property to specify router commands and an <sky-code>extras</sky-code> property to specify <a href="https://angular.io/api/router/NavigationExtras">Angular <sky-code>NavigationExtras</sky-code></a>.
+      Specifies an Angular router link for the action button. This property accepts a custom object with a <sky-code>commands</sky-code> property to specify router commands and an <sky-code>extras</sky-code> property to specify <a href="https://angular.io/api/router/NavigationExtras">Angular <sky-code>NavigationExtras</sky-code></a>. This property passes the <sky-code>commands</sky-code> property to the <a href="https://angular.io/api/router/RouterLink">Angular <sky-code>routerLink</sky-code></a>.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="url"
--- a/package-lock.json
+++ b/package-lock.json
@@ -600,9 +600,9 @@
       "dev": true
     },
     "@skyux-sdk/builder": {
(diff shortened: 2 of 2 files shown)


src/app/components/action-button/index.html
```diff
@@ -72,7 +72,7 @@
       propertyName="route"
       isOptional="true"
     >
-      Specifies an Angular router link for the action button. This property accepts a custom object with a <sky-code>commands</sky-code> property to specify router commands and an <sky-code>extras</sky-code> property to specify <a href="https://angular.io/api/router/NavigationExtras">Angular <sky-code>NavigationExtras</sky-code></a>.
+      Specifies an Angular router link for the action button. This property accepts a custom object with a <sky-code>commands</sky-code> property to specify router commands and an <sky-code>extras</sky-code> property to specify <a href="https://angular.io/api/router/NavigationExtras">Angular <sky-code>NavigationExtras</sky-code></a>. This property passes the <sky-code>commands</sky-code> property to the <a href="https://angular.io/api/router/RouterLink">Angular <sky-code>routerLink</sky-code></a>.
```
Blackbaud-BobbyEarl: I think we could tweak this slightly to better clear up any confusion.

Looking at https://github.com/blackbaud/skyux-layout/blob/7529c88cb598eb420cc81e856eeda4a5ac835c24/src/app/public/modules/action-button/action-button.component.html#L1-L12 we can see the behavior when the `route` property is used.

Specifically, we turn around and pass the following properties to Angular:
-  `permalink.route?.extras?.fragment` to [`fragment`](https://angular.io/api/router/NavigationExtras#fragment)
- `permalink.route?.extras?.queryParams` to [`queryParams`](https://angular.io/api/router/NavigationExtras#queryParams)
- `permalink.route?.extras?.queryParamsHandling` to [`queryParamsHandling`](https://angular.io/api/router/NavigationExtras#queryParamsHandling)
 - `permalink.route?.commands` to [`routerLink`](https://angular.io/api/router/RouterLink)

Blackbaud-DeniseCisneros: I got excited and merged because we Slack chatted about this, sorryyyy! 😳😬  

Would it make sense to add something like, "This property passes the following properties to Angular: " and then include the list above? 

@blackbaud-johnly Thoughts on organization for that?

blackbaud-johnly: That organization makes sense to me, but do we need "to Angular" in the sentence leading into the list @Blackbaud-BobbyEarl?

Blackbaud-BobbyEarl: Probably not necessary as long as we link to the individual parts.  The original confusion around this was that it's in turn handing things over to Angular.

PR #625, opened 2019-08-27
Progress indicator docs redesign

Blackbaud-AdamFunderburk:
Updating to apply docs redesign and add UX content.


src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,266 @@
...
+	
+	<sky-tab
+		tabHeading="Waterfall indicator design">
+        
+	
+    
+    <sky-demo-page-content sectionHeading="Usage">
+	<h3 class="sky-text-success">Use when</h3>
+    
+        <sky-docs-visual-example>
```
Blackbaud-ToddRoberts: Is it worth showing this in a full page context? Seems like we often see examples where the title is top left and the waterfall is centered which looks very weird.

src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,266 @@
...
+        </div>
+        
+        <h3>"Start over"</h3>
+        <div class="sky-text-block">
+            <p>You can include an optional “start over” button that allows the user to erase all of the saved information for the waterfall process. This can return the user to the first step in the waterfall or, if some decision in the task occured before the waterfall, to another previous point in the workflow.</p>
+        </div>
+    </sky-demo-page-content>
+    
+    <sky-demo-page-content sectionHeading="Behavior & states">
+        <div class="sky-text-block">
```
Blackbaud-ToddRoberts: This section has a lot of text, is it possible to trim it down and show examples?

src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,266 @@
...
+					<p>Don't use multiple passive progress indicators at once on items in a list.</p>
+			</sky-docs-visual-example-text>
+			<sky-docs-visual-example-image>
+				<div class="sky-margin-stacked-separate">
+					<img src="~/assets/img/guidelines/progress-indicator/passive-progress-usage-4.png" />
+					<div class="caption caption-danger sky-text-danger">Don't expose multiple passive progress indicators at once.</div>
+				</div>
+                <div>
+					<img src="~/assets/img/guidelines/progress-indicator/passive-progress-usage-5.png" />
+					<div class="caption caption-success sky-text-success">Do roll up steps that are inconsequential into a single status. Use a container like Popover or Flyout to hold the additional details the passive progress indicator provides.</div>
```
Blackbaud-ToddRoberts: Should this show a popover holding the steps?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+    
+    <sky-tab
+      tabHeading="Design">
+    
+      <sky-demo-page-content sectionHeading="Usage">
+        <h3 class="sky-text-success">Use when</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>Use waterfall progress indicators for long or complex setup processes that users typically complete only once. Waterfall indicators are ideal when events outside of the system force users to complete tasks in multiple sittings or when users must enter large amounts of data.</p>
+            <p>Use waterfall progress indicators when processes include at least three steps and the number and order of steps do not change.</p>
```
blackbaud-johnly: We say the best experience requires at least three steps. Should we also recommend an upper limit for the best experience?

Also, should this second paragraph really be treated as part of the first Use when item? It seems like it’s completely distinct from the previous paragraph to me. And it even feels like maybe it should be restated as a Don’t use when item. I think that what we are trying to say is to NOT use waterfalls when the number or order of steps changes, not TO use it whenever processes have stable steps. Same with the number of steps. It seems like maybe we want to recommend NOT to use this pattern for fewer than 3 steps, not to recommend it for process with more than 3 steps.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+            <p>Use waterfall progress indicators when the complexity of steps varies drastically. For example, one step may require a full-screen modal for complex interactions, while another step requires a small modal for a few simple inputs. Waterfall steps can open <a routerLink="/components/wizard">wizards</a> if necessary.</p>
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div class="sky-margin-stacked-separate">
+              <img src="~/assets/img/guidelines/progress-indicator/waterfall-progress-usage-2.png" />
+              <div class="caption caption-success sky-text-success">Do use modals of the appropriate size for the interactions in each waterfall step.</div>
+            </div>
+            <div>
+              <img src="~/assets/img/guidelines/progress-indicator/waterfall-progress-usage-3.png" />
+              <div class="caption caption-danger sky-text-danger">Don't put inputs inside waterfall progress indicators.</div>
```
blackbaud-johnly: Should this be a separate Don’t use when item below instead of a secondary image/caption here? I can definitely see where it’s tangential to the Use when info, but it seems to me like it’s distinct enough to warrant it’s own Don’t use when item below. (Either that or a separate Use when item here that says to use waterfall progress indicators when the actions that users need to perform are going to be performed in inputs on modals that are opened from the waterfall indicator but not by inputs directly on the waterfall indicator.)

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+    
+        <h3 class="sky-text-danger">Don't use when</h3>
```
blackbaud-johnly: Should there be a Don’t use when item calling out not to use the waterfall for processes that users only monitor? And a link to passive indicators? Not sure if this is necessary, but it seemed worth asking.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+        </sky-docs-visual-example>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Anatomy">
+        <sky-docs-anatomy imagePath="~/assets/img/guidelines/progress-indicator/waterfall-progress-anatomy.png">
+          <sky-docs-anatomy-item anatomyNumber="1">Step indicator</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="2">Step label</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="3">Completed step action</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="4">Active step action</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="5">Waterfall title</sky-docs-anatomy-item>
```
blackbaud-johnly: Why is the waterfall title #5? Given that it’s at the top of the page, it seems like it should be #1. But given that it’s included in the Options section, I assume that the problem is that this should actually be marked as optional. Is that correct?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+        
+        <h3>Start over action</h3>
+        <div class="sky-text-block">
+          <p>You can include a start over action to let users erase all saved data from the waterfall process. This returns user to the first step or to an earlier point in the workflow if decisions occured before the waterfall progress indicator.</p>
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Behavior & states">
+        <div class="sky-text-block">
+          <p>Waterfall progress indicators guide users through one active step at a time. When users complete a step, the progress indicator updates its step indicator and step action to the completed state and updates the next step indicator and step action to the active state. Users can review and edit completed steps, but this does not undo the completed state.</p>
```
blackbaud-johnly: If we want to show examples like Todd suggested in the other PR for this, I think we could use an image of side-by-side progress indicators where the current step advances from left to right to show the transition from one step to the next. But I don't know how useful this would be.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+          <p>The final step summarizes the changes and requests user confirmation. You do not need to include all data that users enter in the summary, but confirmation is required. After users complete the final step, replace the waterfall progress indicator with a page that displays the finished or active state of the system. Any page type is allowed, including record pages, work centers, and list pages.</p>
```
blackbaud-johnly: If we want to show examples like Todd suggested, I think we could also use side-by-side images here with a progress indicator on the last step to the left and finished state image on the right. This seems to me like it might be useful.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Content">
```
blackbaud-johnly: I tightened up the paragraphs in the Content section, and one change was to remove the examples in the parentheses in the first paragraph because they were incomplete. None of the examples followed the format that we recommend in that paragraph because they are all missing the direct objects. So I deleted the examples to avoid the confusion of the guideline not matching the examples. If you think it’s important to have examples of step labels, we can add them back, but I would recommend including direct objects in the examples.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+        <div class="sky-text-block">
+          <p>Step labels in waterfall progress indicators start with a number and concisely describe the objective. Use the <b>&lt;verb&gt; &lt;direct object&gt;</b> format. In the label for the last step, describe the cumulative change that occurs at the end of the process.</p>
+          <p>Step action labels closely follow step labels but do not need to be identical. For the first step action label, be generic and conversational (“Get started”), and for completed step actions, use verbs appropriate to post hoc actions (“Change,” “Edit”).</p>
```
blackbaud-johnly: What is the preferred format for action labels? Should they also follow the format that we recommend for the step labels? Users could infer this pattern from the previous paragraph or assume that because we don’t state it explicitly for actions, it doesn’t apply here. I’d remove the ambiguity by being explicit. If this rule applies to both step labels and step actions, I’d make that clear in the first paragraph and then either update or remove the examples for the action labels since they don’t include direct objects. But if it doesn't apply here, I'd explicitly state the recommended format in this paragraph.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,327 @@
...
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Layout">
+        <div class="sky-text-block">
+          <p>Waterfall progress indicators are 400 pixels wide and centered inside the page. The waterfall title is the page title. Do not combine waterfall indicators with other content or actions.</p>
```
blackbaud-johnly: Should the second sentence about how the waterfall title is the page title be in the Content section?

src/app/components/progress-indicator/index.html
Blackbaud-AdamFunderburk: I added an example of a full page context in the Layout section on the page. Think that's good enough?

Blackbaud-ToddRoberts: yup

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: Agreed. I also reworked the description in the Options section as well.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: Done.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I don't think they're necessary here.

src/app/components/progress-indicator/index.html
Blackbaud-AdamFunderburk: Agreed. Updated the visual example.

src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,280 @@
...
+        
+    
+        
+    <sky-demo-page-content sectionHeading="Anatomy">
+        <sky-docs-anatomy imagePath="~/assets/img/guidelines/progress-indicator/waterfall-progress-anatomy.png">
+            <sky-docs-anatomy-item anatomyNumber="1">Step indicator</sky-docs-anatomy-item>
+            <sky-docs-anatomy-item anatomyNumber="2">Step label</sky-docs-anatomy-item>
+            <sky-docs-anatomy-item anatomyNumber="3">Previous step action</sky-docs-anatomy-item>
+            <sky-docs-anatomy-item anatomyNumber="4">Current step action</sky-docs-anatomy-item>
+            <sky-docs-anatomy-item isOptional="true" anatomyNumber="5">Waterfall title</sky-docs-anatomy-item>
```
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: No argument here. This fix is overdue.

src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,280 @@
...
+    <sky-demo-page-example>
+        <sky-progress-indicator-demo>
+        </sky-progress-indicator-demo>
+    </sky-demo-page-example>
+</div>
+    
+<sky-tabset id="content-tabs">
+	
+	<sky-tab
+		tabHeading="Waterfall indicator design">
```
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: Good points, but moot now with the new information architecture.

src/app/components/progress-indicator/index.html
```diff
@@ -1,20 +1,280 @@
...
   pageTitle="Progress Indicator"
-  navTitle="Progress indicator"
-  skyuxModule="SkyProgressIndicatorModule"
-  packageName="@skyux/progress-indicator"
 >
 
   <sky-demo-page-summary>
     <p>
-      The progress indicator component visually represents progress through a series of sequential steps toward a final goal. Active progress indicators represent tasks where user actions are broken into discrete steps, and passive progress indicators represent tasks that concern users but where they do not directly act upon the steps.
+      The progress indicator component shows progress through steps toward a goal. The passive progress indicator represents steps that take place outside of the user’s control. The active progress indicator walks users through a multi-step process, either in the waterfall page (below) or in the <a routerLink="/components/wizard">Wizard</a> component.
```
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: Your suggested rewrite is now the summary on the progress indicator "sub-list". I rewrote the summaries of both child pages to be much more brief. Take a look.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I inserted this point differently into the first DO paragraph, and reiterated it in a DON'T paragraph as well.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I think the demo illustrates this primary behavior well enough. I'm in favor of going without an explicit example to demonstrate the transition, but let's keep an ear out to see if it looks more useful than we think.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I think it's valuable to keep this and its counterpart DO example in tandem. I also think that the two of them in tandem make their point better when leading with the DO example, instead of the DON'T. However, I don't think my opinion here is strong. If you really think we should consider moving this point and swapping the examples, I can be persuaded.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I added an example for this, since it's a mirror of the similar point made on passive indicators.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: Updated.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Makes sense. Thanks.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Resolving as per response on the previous comment.

src/app/components/progress-indicator/index.html
blackbaud-johnly: @Blackbaud-AdamFunderburk, it looks like the first image in the Don't use when section was also updated to a full-page context and is now much larger. None of the other Don't use when images (or Use when images) were updated. Was that change to that one Don't use when image intentional?

src/app/components/progress-indicator/index.html
blackbaud-johnly: Resolving based on other comment threads...

src/app/components/progress-indicator/index.html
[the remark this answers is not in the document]
blackbaud-johnly: Looks good. I'm making one small change locally and will push it up to the PR.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Changes look good. I'm making a couple minor tweaks locally and will push them up to the PR.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: If you think it works better as organized, I think we should stick with it as is. I don't feel strongly about it but wanted to raise the point.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Looks good. I'm gonna push up a minor change to remove the comma...

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Sounds good.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Looks good. I'll push up one minor wording tweak.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Looks good.

src/app/components/progress-indicator/index.html
blackbaud-johnly: Resolving as Adam and I talked about this and the first Use when image was updated...

Issue #563, opened 2019-07-05
Update select field docs to include missing properties


blackbaud-johnly:
In https://developer.blackbaud.com/skyux/components/select-field#skyselectfield-properties, the properties don't match up with the properties in the demo. Based on https://github.com/blackbaud/skyux-select-field/blob/master/src/app/public/modules/select-field/types/select-field.ts, we are missing an `index` property, but that code doesn't include the `description` property from the demo. Apparently, that demo has something to do with another component. (List view grid maybe? I can't remember.) We need to update the documentation to make sure the properties are correctly documented in the correct location.

@Blackbaud-AlexKingman looked into this, so I will coordinate with him on how/where to document the `description` property. (Feel free to add the info directly to the issue if that's easier.)

blackbaud-johnly:
From the email about this:

It looks like `description` gets passed through to the list-view-checklist component, which we do have documented (https://developer.blackbaud.com/skyux/components/list/checklist#list-view-checklist-properties). Now, whether or not that was intentional or just a happy coincidence, I’m not sure. It doesn’t appear that we have any test coverage around this particular property inside of select-field.

Blackbaud-DeniseCisneros:
Completed and merged in https://github.com/blackbaud/skyux2-docs/pull/636

PR #636, opened 2019-09-05
Added description property to SkySelectField properties

Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/563

*Chatted with Alex and we can leave `index` out because it's not for the consumer

### DIFF
--- a/src/app/components/select-field/index.html
+++ b/src/app/components/select-field/index.html
@@ -123,7 +123,7 @@
       propertyName="description"
       isOptional="true"
     >
-    Specifies a description to display below each label when <sky-code>selectMode</sky-code> is set to <sky-code>multiple</sky-code>. This property accepts <sky-code>string</sky-code> values.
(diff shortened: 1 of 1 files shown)


src/app/components/select-field/index.html
```diff
@@ -119,6 +119,12 @@
     <p>
       The <sky-code>SkySelectField</sky-code> type extends the <sky-code>any</sky-code> type and supports values for data objects that you can inject into the select field component to populate the select field modal.
     </p>
+    <sky-demo-page-property
+      propertyName="description"
+      isOptional="true"
+    >
+    Specifies a description to display below each label when <sky-code>selectMode</sky-code> is set to <sky-code>multiple</sky-code>. This property accepts <sky-code>string</sky-code> values.
```
blackbaud-johnly: I'd add "for an item" after "description" and change "each" to "its" ... I mean, "each" makes it sounds like this propert specifies a single description for all items, but I assume this is a per-item description not a generic one, right? ... Other than that, LGTM.

PR #641, opened 2019-09-06
Rework contribution docs edits

blackbaud-johnly:
Updates to https://github.com/blackbaud/skyux2-docs/pull/622 to address https://github.com/blackbaud/skyux2-docs/issues/617.

PR #645, opened 2019-09-10
Updated .travis.yml to use new Chrome addon technique.

Blackbaud-BobbyEarl:
The build has started failing, as seen with https://github.com/blackbaud/skyux2-docs/pull/641.

The code comment in `.travis.yml` suggests looking at http://blog.500tech.com/setting-up-travis-ci-to-run-tests-on-latest-google-chrome-version/, but that link no longer works.

https://docs.travis-ci.com/user/chrome seems to suggest a simpler approach, likely not available when this repo was first created.  This PR is using the documented approach.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,6 @@ dist: trusty
 git:
   depth: 10
 
-services:
-  - xvfb
-
-addons:
-  chrome: stable
-
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,12 @@ dist: trusty
 git:
   depth: 10
 
+services:
+  - xvfb
+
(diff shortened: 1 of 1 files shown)


Blackbaud-BobbyEarl:
Coicindentally, [the build](https://travis-ci.org/blackbaud/skyux2-docs/builds/583349640?utm_source=github_status&utm_medium=notification) seemed to fail with the same error, but because it's being called as an addon, that step didn't fail the build.  

Since there are no tests, not having Chrome didn't seem to have any affect on the entire run either.

Screen Shot 2019-09-10 at 4 43 46 PM

Blackbaud-BobbyEarl:
This is ready again @Blackbaud-SteveBrush .

PR #647, opened 2019-09-12
Character count design docs

Blackbaud-AdamFunderburk:


src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+
+  <sky-demo-page-summary>
+  The character count component is an extension for text input fields with character limits.
+  </sky-demo-page-summary>
+
+  <div
+    class="sky-margin-stacked-content-section"
+  >
+    <sky-demo-page-example>
+      <sky-character-count-demo>
```
blackbaud-johnly: The demo includes name fields with a character limits of 10 characters. Does that seem like a realistic example? I don't think so. Is there a more realistic example that we can use here than first names?

blackbaud-johnly: Also, is there a reason that the reactive form example includes a default value in the field and that the template-driven example does not?

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+              </div>
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+      </sky-demo-page-content>
+
+      <sky-demo-page-content sectionHeading="Anatomy">
+
+        <sky-docs-anatomy
+          imagePath="~/assets/img/guidelines/character-count/character-count-anatomy-1.png"
```
blackbaud-johnly: #1 in this image is in the wrong place. It is more associated with the label than with the input. Looking at the code, I don't think the input is actually part of the component, so I wonder if it should be in this image at all. But if we want a label to show the standard presentation, I think we should still move the #1 to the right so that it is not so closely associated with the label and move it down to overlap with the top border of the input.

#4 is also oddly placed. Why is it overlapping the border? Is this because it is for the red border as well as the status indicator and icon? If so, then I still think it should be moved down a bit to overlap less of the input. But if it is only for the status indicator and message, then I think it should be moved down more.

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+        >
+          <sky-docs-anatomy-item
+            anatomyNumber="1"
+          >
+            Input field
+          </sky-docs-anatomy-item>
+          <sky-docs-anatomy-item
+            anatomyNumber="2"
+          >
+            Character count/limit
```
blackbaud-johnly: What do you think of changing this to `Character count indicator`? If that doesn't work, I'd change `/` to `and`.

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+          </sky-docs-anatomy-item>
+          <sky-docs-anatomy-item
+            anatomyNumber="3"
+          >
+            Danger icon
+          </sky-docs-anatomy-item>
+          <sky-docs-anatomy-item
+            anatomyNumber="4"
+          >
+            <a routerLink="/components/status-indicator"> Status indicator</a> (danger)
```
blackbaud-johnly: It seems weird to only refer to this as `Status indicator` since I think it applies to the red border, shadow, and error message as well, not just the indicator at the far bottom left. If it truly is just the status indicator, then we should move the #4 in the image to the fat bottom left too. But if not, I think we need a different term here. Maybe something more like `Exceeded limit warning` or `Validation for exceeded limit`. Not sure exactly what it should be, but I think the current descriptor is too specific.

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+          <sky-docs-visual-example-text>
+            <p>
+              Don’t use character count on inputs where the character limit is apparent or implied by the data type.
+            </p>
+          </sky-docs-visual-example-text>
+
+          <sky-docs-visual-example-image>
+            <div>
+              <img
+                src="~/assets/img/guidelines/character-count/character-count-usage-2.png"
```
blackbaud-johnly: I think the image here could be slightly confusing. Maybe it doesn't matter, but the indicator shows that the character limit is 10 while the default value in the field has 12 characters counting the parentheses. Should we update the limit to reflect the parentheses? When I test the phone field, it doesn't add the parenthesis by default, so I would think they would count in the character count.

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+        </div>
+
+        <sky-demo-page-component-states>
+          <sky-demo-page-component-state
+            stateName="Empty field"
+            imagePath="~/assets/img/guidelines/character-count/character-count-state-1.png"
+          >
+          </sky-demo-page-component-state>
+          <sky-demo-page-component-state
+            stateName="Adding text"
```
blackbaud-johnly: I like the way that the entries walk through adding text, but are the `Adding text` and `At character limit` items really distinct states? I don't think so, but I'm not sure if it matters either way. Just asking to make sure that if they aren't actually distinct states that we are making a conscious decision to present them as such in order to illustrate how the component works.

src/app/components/character-count/index.html
```diff
@@ -0,0 +1,284 @@
...
+        </sky-demo-page-property>
+        <sky-demo-page-property
+          propertyName="skyCharacterCounterIndicator"
+          isOptional="true"
+        >
+          The character count indicator component. This is the component that displays the count, limit, and over the limit indicator.
+        </sky-demo-page-property>
+        <sky-demo-page-property
+          propertyName="skyCharacterCounterLimit"
+          defaultValue="0"
```
blackbaud-johnly: Does this limit mean that if users don't specify a limit, then the field errors as soon as users enter anything at all? If so, that seems like weird default to me.

Blackbaud-ErikaMcVey: Yeah, it was the default Jack had set but it definitely seems odd. What do we think the default should be, or should it just be a required field?

Blackbaud-ToddRoberts: IMO should be required if using the counter, otherwise there's not much point to using it.

Blackbaud-AdamFunderburk: Agreed, a count limit must be provided to use this. No default will make enough sense to risk encouraging laziness.

src/app/components/character-count/index.html
Blackbaud-AdamFunderburk: Agreed.

src/app/components/character-count/index.html
Blackbaud-AdamFunderburk: Redesigned the anatomy image.

src/app/components/character-count/index.html
Blackbaud-AdamFunderburk: One objective of the Anatomy documentation is to highlight how components build and compose using other components. Being specific about using the actual status indicator class, and not just an error message that looks like one, is part of that objective. I'm inclined to leave it called out that way, and to leave the callout bubble to the right of the text as well (since that's an anatomical part of the status indicator itself).

src/app/components/character-count/index.html
Blackbaud-AdamFunderburk: Updated image, and I made it 13 to include the dash.

src/app/components/character-count/index.html
Blackbaud-AdamFunderburk: Just a conscious decision to show that there's no difference in state when the character limit is matched exactly. It's only when the limit is exceeded that there's a state change. Yes, technically, that means they are not "two different states", but I believe two different examples are needed to make the point completely clear.

src/app/components/character-count/index.html
Blackbaud-ErikaMcVey: No real reason one has the default when the other doesn't other than to show that initializing the field with a value will set the counter correctly. Would you say they should both have a default, or neither?

blackbaud-johnly: I don't have a strong preference. I usually opt for consistency, but in this case, it sounds like it makes sense a default in one and not in the other to highlight the difference in behavior when initializing. I just was checking to make sure this was a conscious decision and not just an accidental inconsistency.

As for the first question? Does anyone have ideas for a more realistic example than a Name field?

src/app/components/character-count/index.html
blackbaud-johnly: So @Blackbaud-AdamFunderburk and @Blackbaud-ToddRoberts, am I correct in assuming that you want us to remove the default on this property and make it required? This would be considered a breaking change, but the component isn't officially released yet so I don't think that matters. Besides, it's highly unlikely that even if anyone was already using it that they would be using the component with the default value of 0. @Blackbaud-SteveBrush, do you have any opinion on this?

src/app/components/character-count/index.html
blackbaud-johnly: Just double-checking to see if the Name example is a legit use case. If so, we can leave it as is. If not, maybe we can tweak it to something more realistic.

src/app/components/character-count/index.html
blackbaud-johnly: I updated this to hide the default and marked the property as required after discussing this with the team.

Issue #466, opened 2019-05-13
Improve documentation on implications of using dataProvider on a list component


Blackbaud-AlexKingman:
https://developer.blackbaud.com/skyux/components/list/overview

Using the `dataProvider` inherently comes with some significant implementation differences. The docs don't necessarily make this very clear.

Important notes:

- Filtering / searching is now the responsibility of the consumer
- If consumer wants to enable multiselect toolbar, they'll also have to add logic to handle the "show only selected" checkbox, as this is also a filter.
- We should flesh out the demo to better demonstrate these points ☝️

Blackbaud-DeniseCisneros:
Per Slack convo, would probably also make sense to add best practice info based on different use-cases

PR #648, opened 2019-09-13
Added info about filtering and searching for dataProvider

Blackbaud-DeniseCisneros:
Addresses https://github.com/blackbaud/skyux2-docs/issues/466

@Blackbaud-AlexKingman, will you provide more insight on how the filtering is done and then I can add that?


src/app/components/list/overview/index.html
```diff
@@ -32,7 +32,7 @@
...
+    Specifies a data provider to obtain the data to display. The list component requires this property or the <sky-code>data</sky-code> property. Acceptable values: <sky-code>extends ListDataProvider</sky-code>. To extend ListDataProvider, provide a <sky-code>get</sky-code> function that accepts a <sky-code>ListDataRequestModel</sky-code> as an argument and returns an <sky-code>Observable ListDataResponseModel </sky-code>, and a <sky-code>count</sky-code> function that returns an <sky-code>Observable number</sky-code> that indicates how many total records are in the list. You must control the filtering and searching when you opt in to the <sky-code>dataProvider</sky-code>.
```
blackbaud-johnly: A few things...

Is this new sentence strong enough? I mean, what do we mean by "you must control"? Does that adequately indicate that the user (if it is in fact the user) is taking responsibility? I wonder if it would be better to say something more like "For lists that use `dataProvider` instead of `data`, the consumer must take responsibility for filtering and searching." Not really sure; that's just my initial question after looking at this and the first bullet of the description at https://github.com/blackbaud/skyux2-docs/issues/466.

And speaking of https://github.com/blackbaud/skyux2-docs/issues/466, this seems to correlate to the first bullet in that description. Do we need to include something for the second bullet as well?

And do we need to create a separate issue for updating the demo?

Blackbaud-AlexKingman:
@Blackbaud-DeniseCisneros After thinking about this for a while, I think it would be really difficult to explain in words how to build custom filters. I wonder if we should update the demo to make that more clear (right now it doesn't show filters), and then refer to the demo in the docs. Something like, `Refer to the demo below to see how to write your own custom searching, filters, and paging.` What do you think?

Blackbaud-DeniseCisneros:
Do you think we should keep the line I added about the user needing to control the filtering and searching and then refer the user to the demo?

Blackbaud-AlexKingman:
@Blackbaud-DeniseCisneros Yeah I like the added sentence. It can totally be a lead-in to referring to the demo.

FWIW - its not just limited to filtering and searching. Here's a full list of the properties on `ListDataRequestModel` they'll now have to manage:
```
  public filters: ListFilterModel[];
  public pageSize: number;
  public pageNumber: number;
  public search: ListSearchModel;
  public sort: ListSortModel;
  public isToolbarDisabled: boolean = false;
```

I don't necessarily think we need to include all these, but it might be a good idea to give them an idea that managing all properties of `ListDataRequestModel` is now the responsibility of the consumer.

Blackbaud-DeniseCisneros:
I tweaked the wording some based on what both of you all said. I didn't list everything they needed to manage but instead specifically said users need to manage all `ListDataRequestModel` properties.

Alex, thoughts on creating a separate issue for the demo?

Blackbaud-AlexKingman:
LGTM!

I added a new issue for the demo: #653 

Once we get this build passing, we can ship it!

Blackbaud-DeniseCisneros:
John, I can pick up the related issue you created for the additional docs. Thanks for creating that!

Blackbaud-BobbyEarl:
Closing + reopening to trigger CI.

PR #649, opened 2019-09-13
Single file attach docs

Blackbaud-ToddRoberts:


Blackbaud-AdamFunderburk:
- Can some of the summary paragraphs be moved to the Development tab, so we can have a shorter summary description of the component(s)? Some of the content in the current summary is pretty in-the-weeds technical considerations.

 - Progress indicator's revisions also use two tabs for design content, but title them "____ design". These are titled "Design - _____". Want to reconcile them to the former or latter?

 - Do these modes need a Behavior section, spelling out how they transform when files are uploaded? The reader can work it out using the images in the States table, but writing it out can add a lot of details that might otherwise get lost (like the previews push the rest of the form content down when they appear), and it's also better for accessibility.

 - I think the "Attach from a URL" piece of the File attachment anatomy might need to be declared optional.

 - Less certain, but should the image preview for the Single file attachment be considered "optional" since it only appears in the case of image files? Probably not, since it's not really at the author's discretion...

 - Worth it to include a DON'T image example for the case of using File attachment for a single local file?

 - There are two cases where there is a guideline with no accompanying image, and it's using the text-block with a wider width than the visual-example-text. Do you think the different paragraph widths is trouble? Can the sentence just be combined with the example underneath it?

Blackbaud-ToddRoberts:
> * Can some of the summary paragraphs be moved to the Development tab, so we can have a shorter summary description of the component(s)? Some of the content in the current summary is pretty in-the-weeds technical considerations.

Moved, @blackbaud-johnly or @Blackbaud-DeniseCisneros may want to take a look and see if it needs to be cleaned up now that it's on that tab.

> * Progress indicator's revisions also use two tabs for design content, but title them "____ design". These are titled "Design - _____". Want to reconcile them to the former or latter?

My preference is "Design - " since the design/dev distinction seems more important to me as a first evaluation when looking for relevant content. 

> * Do these modes need a Behavior section, spelling out how they transform when files are uploaded? The reader can work it out using the images in the States table, but writing it out can add a lot of details that might otherwise get lost (like the previews push the rest of the form content down when they appear), and it's also better for accessibility.

Added
 
> * I think the "Attach from a URL" piece of the File attachment anatomy might need to be declared optional.

Updated

> * Less certain, but should the image preview for the Single file attachment be considered "optional" since it only appears in the case of image files? Probably not, since it's not really at the author's discretion...

Added a note in the behavior section to call this out.

> * Worth it to include a DON'T image example for the case of using File attachment for a single local file?

I thought about it but couldn't come up with a good way to show something that's not possible.

> * There are two cases where there is a guideline with no accompanying image, and it's using the text-block with a wider width than the visual-example-text. Do you think the different paragraph widths is trouble? Can the sentence just be combined with the example underneath it?

The different paragraph size doesn't bother me too much. I _think_ it's better than combining with the next paragraphs since the content isn't really related.

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+				<div>
+					<img src="~/assets/img/guidelines/single-file-attach/dont-use-multi.png" />
+					<div class="caption caption-danger sky-text-danger">Don't use single file attachment if the user might upload multiple files.</div>
+        </div>
+			</sky-docs-visual-example-image>
+		</sky-docs-visual-example>
+</sky-demo-page-content>
+  
+  <sky-demo-page-content sectionHeading="Anatomy">
+		<sky-docs-anatomy imagePath="~/assets/img/guidelines/single-file-attach/anatomy.png">
```
blackbaud-johnly: Should the button in the second half of the image with the Replace file button also have a green 1 on it? I think so unless we consider that a totally different button, and if that's the case, then I think it should have a separate entry in the anatomy list.

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+
+    <h3 class="sky-text-danger">Don't use when</h3>
+      <p class="sky-text-block">
+        Don't use file attachment if the user needs to upload one local file. Use single file attachment instead.
+      </p>
+  </sky-demo-page-content>
+
+    <sky-demo-page-content sectionHeading="Anatomy">
+      <sky-docs-anatomy imagePath="~/assets/img/guidelines/file-attach/anatomy.png">
+        <sky-docs-anatomy-item anatomyNumber="1">Browse or drop local file</sky-docs-anatomy-item>
```
blackbaud-johnly: This item and #6 both feel like they are missing a noun. All the other items in the list are nouns that correspond to UI items. But #1 and #6 are the actions that the UI items enable. What do you think of changing these to something like `Upload file option` and `Attach URL option`?

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+        <sky-docs-anatomy-item anatomyNumber="2">File name</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="3">File size</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="4">File preview</sky-docs-anatomy-item>
```
blackbaud-johnly: Should this be `File preview` or `Image preview`? We say `File preview` here but `Image preview in the anatomy list for single file attachment.

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+        <sky-docs-anatomy-item anatomyNumber="5">Delete button</sky-docs-anatomy-item>
```
blackbaud-johnly: Why does the button here look different than the delete button for single file attachment? This one is a black icon with a border, while that one is a light grey icon with no border. Seems like they should probably look the same since they are serving the same purpose.

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+        <sky-docs-anatomy-item anatomyNumber="6" isOptional="true">Attach from a URL</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="7" isOptional="true">Additional metadata</sky-docs-anatomy-item>
```
blackbaud-johnly: I'd delete `Additional` since that implies that there is already some metadata. Unless I'm missing something, I don't think that's the case.

src/app/components/fileattachments/index.html
```diff
@@ -1,25 +1,213 @@
...
+    </sky-demo-page-content>
+    
+    <sky-demo-page-content sectionHeading="Behavior & states">
+      <h3>File information</h3>
+      <sky-docs-visual-example>
+        <sky-docs-visual-example-text>
+            <p>When the user attaches a file, summary information for that file appears in a card below the file attach input. By default the summary information includes file name, a delete button, and if the file is from a local drive also includes file size and a file thumbnail. You can optionally include inputs for the user to add metadata about the file.</p>
+        </sky-docs-visual-example-text>
+        <sky-docs-visual-example-image>
+          <img src="~/assets/img/guidelines/file-attach/behavior-preview.png" />
```
blackbaud-johnly: Why does this image user borders when the other don't?

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: single file attach does not show a preview for non-image formats

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: yup, added

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: updated

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: the inline/borderless button doesn't work on the gray bg of the file attach file info panel, but is the correct usage for single file attach

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: to more clearly demarcate the before/after states, since the arrow is +/- the same color as a lot of the UI it seemed confusing without the borders.

src/app/components/fileattachments/index.html
Blackbaud-ToddRoberts: updated

src/app/components/fileattachments/index.html
blackbaud-johnly: Gotcha. Thanks.

src/app/components/fileattachments/index.html
blackbaud-johnly: Sounds good. Thanks.

src/app/components/file-attachments/single-file-attachment/index.html
```diff
@@ -77,7 +77,8 @@ <h3>States</h3>
       <sky-demo-page-component-state stateName="Base" imagePath="~/assets/img/guidelines/single-file-attach/state-base.png"></sky-demo-page-component-state>
       <sky-demo-page-component-state stateName="Attached image" imagePath="~/assets/img/guidelines/single-file-attach/state-attached-image.png"></sky-demo-page-component-state>
       <sky-demo-page-component-state stateName="Attached non-image" imagePath="~/assets/img/guidelines/single-file-attach/state-attached-non-image.png"></sky-demo-page-component-state>
-      <sky-demo-page-component-state stateName="Error" imagePath="~/assets/img/guidelines/single-file-attach/state-error.png"></sky-demo-page-component-state>
+      <sky-demo-page-component-state stateName="Missing required file error" imagePath="~/assets/img/guidelines/single-file-attach/state-error-missing.png"></sky-demo-page-component-state>
```
blackbaud-johnly: This is the only image in this table where the label appears above the input. Should we crop it out of the image for consistency?

blackbaud-johnly: Gotcha. Makes sense. Thanks!

Issue #284, opened 2017-09-08
Document SCSS code style guidelines for contributing components to SKY UX


blackbaud-johnly:
Document SCSS files and direct developers to the correct mixins/classes/variables. These should already be documented but are spread out in different areas of the docs site and in the UX code itself. So developers might not know where to look for this info, or they might not even know that they are expected to use those styles instead of hard coding values. This documentation would most likely consist mostly of pointers to the areas where developers can find the appropriate styles, plus a little bit of explanation.

Blackbaud-DeniseCisneros:
Checked in with @Blackbaud-PaulCrowder. Placing this on hold due to current theming work.

Blackbaud-SteveBrush:
Closing issue due to inactivity. If you believe this should still be addressed, please file a new issue and include any additional context and justification. Thank you.

PR #661, opened 2019-09-18
[HOLD] Added blurb about SCSS file

Blackbaud-DeniseCisneros:
Added blurb about SCSS file to address https://github.com/blackbaud/skyux2-docs/issues/284.


Blackbaud-DeniseCisneros:
Plan to iterate on wording but wanted to see what you thought of the first pass of this.

src/app/learn/reference/naming-and-placement/index.html
```diff
@@ -48,6 +48,11 @@
         <sky-code>src/assets/locales</sky-code> folder — To <a routerLink="/learn/get-started/advanced/localization">specify localization strings</a> to reuse throughout your SPA, declare the strings in the <sky-code>resources_en_US.json</sky-code> file in <sky-code>src/assets/locales</sky-code>. To reference the localization strings in your SPA, use the <sky-code>skyAppResources</sky-code> pipe or the <sky-code>SkyAppResourcesService</sky-code> service.
       </p>
     </li>
+    <li>
+      <p>
+        <sky-code>SCSS</sky-code> file — This file contains the import that includes design tokens for SKY UX building styles. You can also add custom styles to this file because SCSS is a superset of CSS3 syntax. For more information, see the <a href="http://sass-lang.com/">SASS website</a>.
```
blackbaud-johnly: If we are referring generically to SCSS, it doesn't need to be a code snippet. We use that style for folders and file names, but you can see a couple entries up that when we refer generically to SCSS, we don't use a code snippet.

The placement of this item in the list seems off to me. Looking in the docs repo, I see a `styles.scss` file in the `src/app` folder, so I assume that's where this file is supposed to be placed. Maybe we should move this up in the list after the `src/app/lib` and before `src/assets` to indicate that. What do you think?

Are we referring to a template file here or something that users need to add? I assume the latter. If the former, we need an entry in the template items topic at https://developer.blackbaud.com/skyux/learn/reference/template-items. But I'm pretty sure this is a file that users need to create. But "This file contains..." makes it sound like this is an existing file. I think this should probably be something more like "You can create an SCSS file to..."

I'm not sure what we mean when we say "contains the import that includes design tokens for SKY UX building styles." that sounds very circular to me. Can we just collapse this to "import design tokens for SKY UX building styles"? Or am I missing some nuance here? Also, can we clarify what we mean by design tokens for SKY UX building styles? What exactly are we telling users to import in this file? Would it be useful to provide more context, and can we provide links and/or examples? At the top of the `styles.scss` file in the docs repo, I see the line `@import '~@blackbaud/skyux-design-tokens/scss/mixins';`, so I assume this is at least part of what we are talking about. The issue that these docs references directing developers to the correct mixins/classes/variables, so I'm not sure if additional imports are sometimes necessary. I don't see any others in that `styles.scss` file, so maybe it's just that one import. I think we should bounce these questions off one of the developers on the team.

In the second sentence, is it important to specify CSS3 instead of just CSS?

Finally, it's minor, but on the website that we link to "Sass" is not all caps.

src/app/learn/reference/naming-and-placement/index.html
```diff
@@ -38,6 +38,11 @@
         <sky-code>src/app/lib</sky-code> folder — This folder is the recommended location for files to exclude from code coverage. SKY UX's code coverage instrumentation ignores all files in the <sky-code>lib</sky-code> folder.
       </p>
     </li>
+    <li>
+      <p>
+        SCSS file — To add custom styles, you create the SCSS file and add <sky-code>@import '~@skyux/theme/scss/mixins';</sky-code> to reference variables or mixins. For more information, see the <a href="http://sass-lang.com/">Sass website</a>.
```
blackbaud-johnly: Where do users create this file? I think we should say where to create the file. My original feedback suggested moving this item here because I thought it goes in the `src/app` folder. I think we should check with a developer to confirm the location and specify it in the description.

Also, where do consumers add `@import '~@skyux/theme/scss/mixins';`? Is it in this SCSS file? I assume so from how this is written, but I want to make sure.

Also, The last sentence seems entirely out of context. I'd add `SCSS is a superset of CSS3 syntax.` befroe the last sentence to provide context.

Blackbaud-DeniseCisneros:
After chatting with the team, this content should live on a brand new page instead of being added to an existing page/section. Closing in favor of the newer PR that will get created.

PR #670, opened 2019-09-23
Select field custom picker edits

blackbaud-johnly:
Suggested edits for the new custom picker docs.


src/app/components/select-field/index.html
```diff
@@ -164,33 +164,34 @@
   <sky-demo-page-properties sectionHeading="SkySelectFieldPickerContext properties">
     <sky-demo-page-property
       propertyName="data"
-      isOptional="true"
+      isOptional="false"
```
Blackbaud-PaulCrowder: This property is set by the select field itself and passed to the custom picker, as opposed to a field the consumer of the component sets. Therefore a "required" property doesn't really make sense on any of these properties.

src/app/components/select-field/index.html
```diff
@@ -164,33 +164,34 @@
...
     >
-      An array of <a stacheRouterLink="." fragment="skyselectfield-properties">The <sky-code>SkySelectField</sky-code> objects</a> provided to the select field's <sky-code>data</sky-code> property. Use this property to populate your custom picker with options for the user to select.
+      Specifies an array <a stacheRouterLink="." fragment="skyselectfield-properties">of <sky-code>SkySelectField</sky-code> objects</a> and provides it to the select field's <sky-code>data</sky-code> property. This populates the custom picker with items for users to select.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="headingText"
       isOptional="true"
     >
-      The text provided to the select field's <sky-code>headingText</sky-code> property. If your custom picker is a modal, use this property for the modal's title.
+      Specifies a header for the select field component's modal and provides it to the select field's <sky-code>pickerHeading</sky-code> property. This property accepts <sky-code>string</sky-code> values.
```
Blackbaud-PaulCrowder: The rewording here is also misleading. This value is passed from the select field to the custom picker, not from the consumer to the component. It's up to the author of the custom picker to use this value to set the picker heading if applicable.

src/app/components/select-field/index.html
```diff
@@ -164,33 +164,34 @@
...
     </sky-demo-page-property>
     <sky-demo-page-property
-      propertyName="selectedValue"
+      propertyName="selectMode"
       isOptional="true"
     >
-      The select field's current value. If the select field's <sky-code>selectMode</sky-code> is set to <sky-code>single</sky-code>, this will be a <sky-code>SkySelectField</sky-code> object; otherwise it will be an array of <sky-code>SkySelectField</sky-code> objects. Use this property to select the corresponding items in the custom picker.
+      Specifies the selection mode that determines whether users can select one item or multiple items and provides the value to the select field's <sky-code>selectMode</sky-code> property.
```
Blackbaud-PaulCrowder: Another place where the rewording is misleading. This value is provided by the select field to the custom picker, not by the consumer to the select field. The rewording also obscures the fact that the value provided by the consumer to the select field is what is passed through to the custom picker.

src/app/components/select-field/index.html
```diff
@@ -164,33 +164,34 @@
...
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="showAddNewRecordButton"
       isOptional="true"
+      defaultValue="false"
     >
-      The value provided to the select field's <sky-code>showAddNewRecordButton</sky-code> property. Use this property to determine whether a new record can be added from the custom picker when the desired value is not in the list of options.
+      Indicates whether to display a button for users to add new items directly from the select field modal and provides the value to the select field's <sky-code>showAddNewRecordButton</sky-code> property. This property accepts <sky-code>boolean</sky-code> values.
```
Blackbaud-PaulCrowder: The rewording to "provides the value to the select field's `showAddNewRecordButton`" is problematic here for the same reasons as the ones stated above.

src/app/components/select-field/index.html
blackbaud-johnly: OK. I'll change it back. I just saw in https://github.com/blackbaud/skyux-select-field/blob/master/src/app/public/modules/select-field/select-field-picker-context.ts that this was the only property without the question mark or a default, so I thought that meant it was required.

src/app/components/select-field/index.html
```diff
@@ -164,33 +164,34 @@
   <sky-demo-page-properties sectionHeading="SkySelectFieldPickerContext properties">
     <sky-demo-page-property
       propertyName="data"
-      isOptional="true"
+      isOptional="false"
     >
-      An array of <a stacheRouterLink="." fragment="skyselectfield-properties">The <sky-code>SkySelectField</sky-code> objects</a> provided to the select field's <sky-code>data</sky-code> property. Use this property to populate your custom picker with options for the user to select.
+      Specifies an array <a stacheRouterLink="." fragment="skyselectfield-properties">of <sky-code>SkySelectField</sky-code> objects</a> and provides it to the select field's <sky-code>data</sky-code> property. This populates the custom picker with items for users to select.
```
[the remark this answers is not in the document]
blackbaud-johnly: I reworked the updates to try to address your feedback in the four places where the changes were misleading. Hopefully this works better, but if not, let me know and I'll work on it some more.

src/app/components/select-field/index.html
```diff
@@ -166,31 +166,32 @@
...
+      Receives the select field's current value. If the select field's <sky-code>selectMode</sky-code> is set to <sky-code>single</sky-code>, then this value is <a stacheRouterLink="." fragment="skyselectfield-properties">a <sky-code>SkySelectField</sky-code> object</a>. If the select field's <sky-code>selectMode</sky-code> is set to <sky-code>multiple</sky-code>, then this value is an array <a stacheRouterLink="." fragment="skyselectfield-properties">of <sky-code>SkySelectField</sky-code> objects</a>. Use this property to select the corresponding items in the custom picker.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="selectMode"
       isOptional="true"
     >
-      The value provided to the select field's <sky-code>selectMode</sky-code> property. Use this property to determine whether the custom picker should allow more than one value to be selected.
+      Receives a <sky-code>SkySelectFieldSelectMode</sky-code> value from the select field's <sky-code>selectMode</sky-code> property. Use this property to specify the selection mode that determines whether users can select one item or multiple items.
```
Blackbaud-PaulCrowder: The change from "use this property to determine whether" to "use this property to specify" alters the meaning of the property. The consumer should be looking at the property's value to determine the action to take, but the edit makes it sound like the consumer should be specifying the value. How about this?

"The <sky-code>SkySelectFieldSelectMode</sky-code> value from the select field's <sky-code>selectMode</sky-code> property. Use this property to determine whether users can select one item or multiple items."

src/app/components/select-field/index.html
```diff
@@ -166,31 +166,32 @@
...
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="showAddNewRecordButton"
       isOptional="true"
+      defaultValue="false"
     >
-      The value provided to the select field's <sky-code>showAddNewRecordButton</sky-code> property. Use this property to determine whether a new record can be added from the custom picker when the desired value is not in the list of options.
+      Receives a <sky-code>boolean</sky-code> value from the select field's <sky-code>showAddNewRecordButton</sky-code> property. Use this property to indicate whether to display a button for users to add new items directly from the custom picker's modal.
```
Blackbaud-PaulCrowder: Similar problem here as above with "determine" vs. "indicate." How about this?

"The <sky-code>boolean</sky-code> value from the select field's <sky-code>showAddNewRecordButton</sky-code> property. Use this property to determine whether to display a button for users to add new items directly from the custom picker."

src/app/components/select-field/index.html
```diff
@@ -166,31 +166,32 @@
...
     >
-      An array of <a stacheRouterLink="." fragment="skyselectfield-properties">The <sky-code>SkySelectField</sky-code> objects</a> provided to the select field's <sky-code>data</sky-code> property. Use this property to populate your custom picker with options for the user to select.
+      Receives an array <a stacheRouterLink="." fragment="skyselectfield-properties">of <sky-code>SkySelectField</sky-code> objects</a> from the select field's <sky-code>data</sky-code> property. Use this property to populate the custom picker with items for users to select.
     </sky-demo-page-property>
     <sky-demo-page-property
       propertyName="headingText"
       isOptional="true"
     >
-      The text provided to the select field's <sky-code>headingText</sky-code> property. If your custom picker is a modal, use this property for the modal's title.
+      Receives text from the select field's <sky-code>pickerHeading</sky-code> property. Use this property to display a header in the custom picker's modal.
```
Blackbaud-PaulCrowder: Not all custom pickers must be modal, so I think it's still worth mentioning the "if your custom picker is a modal" part.

src/app/components/select-field/index.html
```diff
@@ -166,31 +166,32 @@
       propertyName="data"
       isOptional="true"
     >
-      An array of <a stacheRouterLink="." fragment="skyselectfield-properties">The <sky-code>SkySelectField</sky-code> objects</a> provided to the select field's <sky-code>data</sky-code> property. Use this property to populate your custom picker with options for the user to select.
+      Receives an array <a stacheRouterLink="." fragment="skyselectfield-properties">of <sky-code>SkySelectField</sky-code> objects</a> from the select field's <sky-code>data</sky-code> property. Use this property to populate the custom picker with items for users to select.
```
Blackbaud-PaulCrowder: I'm not sure about the "Receives" wording. It's not a term typically used by developers when describing properties on objects. I think a simple description of the property would be sufficient here (i.e. "An array of SkySelectField objects...").

blackbaud-johnly:
I updated the entries based on your comments and our discussion. Hopefully these look good now, but let me know if you see anything else. Thanks.

PR #676, opened 2019-09-30
Edits to base property docs

blackbaud-johnly:


src/app/learn/get-started/basics/build-publish/index.html
```diff
@@ -5,10 +5,10 @@
...
+      Before you build and publish your SPA, you must choose a web server to host your site. We recommend <a href="https://pages.github.com/">GitHub Pages</a>, a free tool suitable for front-end projects.
     </p>
     <sky-alert alertType="info">
-      Blackbaud developers should create repos and provision applications through the SKY UX App Management Portal. Documentation on the SKY UX App Management Portal is available on <a href="https://host.nxt.blackbaud.com/engineering-system-docs/learn/spa/spa-app-portal">the Engineering System website</a>. This site is for Blackbaud internal use only and is not available to external developers.
+      Blackbaud developers create repos and provision applications through the SKY UX App Management Portal. Documentation on the SKY UX App Management Portal is available on <a href="https://host.nxt.blackbaud.com/engineering-system-docs/learn/spa/spa-app-portal">the Engineering System website</a>. This site is for Blackbaud internal use only and is not available to external developers.
```
Blackbaud-BobbyEarl: I realize it was already there, can we change that to `https://docs.blackbaud.com/engineering-system-docs/learn/spa/spa-app-portal`?

src/app/learn/get-started/basics/build-publish/index.html
```diff
@@ -20,17 +20,14 @@
...
       </p>
-      <p>
-          For example, if you provided <sky-code>my-spa</sky-code> after running <sky-code>skyux new</sky-code>, change the <sky-code>dist/</sky-code> folder to <sky-code>my-spa/</sky-code>.
-      </p>
-      <sky-alert alertType="info">If you would like to host your SPA at the root, without having to use a subfolder, you can set the <a routerLink="/learn/reference/configuration/apply-appsettings">base property</a> in <sky-code>skyuxconfig.json</sky-code> to <sky-code>/</sky-code>.</sky-alert>
+      <sky-alert alertType="info">To host your SPA at the root without a subfolder, set the <sky-code>base</sky-code> property to <sky-code>/</sky-code> in <a routerLink="/learn/reference/configuration" fragment="app">the <sky-code>app</sky-code> configuration option</a> in <sky-code>skyuxconfig.json</sky-code>.</sky-alert>
```
Blackbaud-BobbyEarl: Can we say section instead of option?

blackbaud-johnly: We should if you think that's clearer or more accurate. ... But FWIW, we use "option" elsewhere, so this would be a tweak to the terminology we currently use. For example, we use "option" in the warning at the end of this topic. But if "section is more accurate, I can change both and make a note to update it elsewhere too.

src/app/learn/get-started/basics/build-publish/index.html
blackbaud-johnly: I changed "host" to "docs" in the URL.

src/app/learn/get-started/basics/build-publish/index.html
Blackbaud-BobbyEarl: I think the reason those feel different are because of the type of values they support.  

In the case of `app`, it's only valid values are child properties/options, which is why I suggested "section."

In the case of `base`, it's only valid values are strings.

blackbaud-johnly: I updated it to "section."

Issue #654, opened 2019-09-18
Document properties of ListDataRequestModel in the list component docs


blackbaud-johnly:
The properties in `ListDataRequestModel` are public, and in https://github.com/blackbaud/skyux2-docs/pull/648 we call out that the consumer is responsible for those properties if they use `dataProvider` instead of `data`. We need to add a new section to document these properties, and we should link to it form the `dataProvider` property.

Blackbaud-DeniseCisneros:
Merged and closed in https://github.com/blackbaud/skyux2-docs/pull/692

PR #692, opened 2019-10-10
Added additional properties for ListDataRequestModel

Blackbaud-DeniseCisneros:
_Finally_ getting this thing opened! Addresses https://github.com/blackbaud/skyux2-docs/issues/654

I would definitely appreciate some extra attention on this PR. Please and thanks!


src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+      propertyName="defaultValue"
+      isOptional="true"
+    >
+    Specifies the default value of the filter. If the value of the filter is set to the default value, then the filter is not applied.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="filterFunction"
+      isOptional="true"
+    >
+    Specifies the function to apply to determine whether an item is filtered. This property is required when using in memory data provider. Acceptable values: <sky-code>(item: ListItemModel, filter: any) => boolean</sky-code>. This property accepts <sky-code>boolean</sky-code> values.
```
blackbaud-johnly: Except for the last sentence, this description is the same as the description of the `filters` property of `ListDataRequestModel`. Is that correct? I'm pretty sure we don't want the last sentence here. And looking at the code, I'm not sure if the Acceptable values sentence in the `filters` description is there. It looks like the last sentence there should be "This property accepts a `ListFilterModel` value" but you might want to check with a dev.

If it's correct that the description here should be the same as for `filters`, I'd recommend the same changes here that I suggested there.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="filterFunction"
+      isOptional="true"
+    >
+    Specifies the function to apply to determine whether an item is filtered. This property is required when using in memory data provider. Acceptable values: <sky-code>(item: ListItemModel, filter: any) => boolean</sky-code>. This property accepts <sky-code>boolean</sky-code> values.    
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSearchModel properties">
```
blackbaud-johnly: We should organize the properties in alphabetical order.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+      propertyName="searchText"
+      isOptional="true"
+    >
+    Specifies the text to search. This property accepts <sky-code>string</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="functions"
+      isOptional="true"
+    >
+    Specifies an array of functions that should return a <sky-code>boolean</sky-code> value of <sky-code>true</sky-code> when the search is successful. Acceptable values: <sky-code>(data: any, searchText: string) => boolean></sky-code>.
```
blackbaud-johnly: I'd delete "should."

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies an array of <sky-code>string</sky-code> types which run against search fields. The <sky-code>string</sky-code> values should match the data's <sky-code>field</sky-code> value.
```
blackbaud-johnly: I'm not following this description. What does it mean to run types against search fields? And what `field` values are we talking about? We format `field` as a code snippet, which suggests were are referring to a specific property, but I don't see any `field` property anywhere else on this page. I suspect this description should be something like "Specifies the types to search. The types correspond to `field` values that you specify with the `data` property. This property accepts an array of `string` values." (where "the `data` property" could link to the List properties section.) But that said, I'm really not entirely sure if my understanding is correct here, and we need to sort this out with a dev.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSortModel properties">
```
blackbaud-johnly: We should use alphabetical order for the items in this list.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    <sky-demo-page-property
+      propertyName="available"
+      isOptional="true"
+    >
+    Specifies the list of avilable views to sort. Acceptables values: <sky-code>Array of ListSortLabelModel</sky-code>.
```
blackbaud-johnly: "available" is misspelled. ... I'd change the last sentence to "This property accepts an array of `ListSortLabelModel` objects." And should we have another table with the properties of `ListSortLabelModel`? It seems like we should based on https://github.com/blackbaud/skyux-list-builder/blob/master/src/app/public/modules/list/state/sort/sort.model.ts. It looks to me like there are 5 properties there to document. (FWIW, I wouldn't necessarily worry about that for this PR. I'd just confirm with dev whether we should document those properties, and, if so, we can either handle it here or create a new docs issue to circle back to that.)

blackbaud-johnly: Also, what does "views" mean here? What exactly is being sorted?

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="global"
+      isOptional="true"
+    >
+    Specifies all views to sort. Acceptable values: <sky-code>Array of ListSortLabelModel</sky-code>.
```
blackbaud-johnly: Same recommendation for the second sentence here as in the previous description. And same question about "views."

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies the fields to sort. Acceptable values: <sky-code>Array of ListSortFieldSelectorModel</sky-code>.
```
blackbaud-johnly: Same basic concern here as in the previous 2 items. I'd change the last sentence to "This property accepts an array of ListSortFieldSelectorModel objects." And should we have another table with the properties of ListSortLabelModel? It's not actually in this module, but it seems like we should have something based on https://github.com/blackbaud/skyux-list-builder-common/blob/master/src/app/public/state/sort/field-selector.model.ts. It looks to me like there are 2 properties there to document.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
+  <sky-demo-page-properties sectionHeading="ListDataRequestModel properties">
```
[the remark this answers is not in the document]
Blackbaud-DeniseCisneros: After a chat with Trevor and Alex, I thought we landed on not including `isToolbarDisabled`, but in re-reading the slack conversation, I'm not sure if they meant they wouldn't include it in the demo. @Blackbaud-AlexKingman, wanna chime in on whether or not we should doc `isToolbarDisabled` after all?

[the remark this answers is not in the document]
Blackbaud-AlexKingman: Yeah, I don't think we'll include it in the demo but we could certainly include it here. It simply disables the search bar and the filter button. Accepts booleans. Default is false.

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+    <sky-demo-page-property
+      propertyName="filters"
+      isOptional="true"
+    >
+    Specifies the function to apply to determine whether an item is filtered. This property is required when using in memory data provider. Acceptable values: <sky-code>(item: ListFilterModel, filterValue: any) => boolean</sky-code>.
```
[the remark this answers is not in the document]
Blackbaud-AlexKingman: 'items' sounds fine to me.

Actually, this property accepts an array of type `ListFilterModel`. See this line here:
https://github.com/blackbaud/skyux-list-builder/blob/5e89a6f68a926949e9c784820adbc30e0102380c/src/app/public/modules/list/list-data-request.model.ts#L6

src/app/components/list/overview/index.html
```diff
@@ -60,6 +60,120 @@
...
+      propertyName="pageSize"
+      isOptional="true"
+    >
+    Specifies the number of items to display per page. This property accepts <sky-code>number</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="search"
+      isOptional="true"
+    >
+    Specifies a function to dynamically manage the data source when users change the text in the list field. The search function must return an array or a promise of an array. The <sky-code>search</sky-code> property is particularly useful when the data source does not live in the source code. Acceptable value: <sky-code>(searchText: string, data?: any[]) => any[] | Promise&lt;any[]&gt;</sky-code>.
```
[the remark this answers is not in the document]
Blackbaud-AlexKingman: Agree on using `ListSearchModel` as the acceptable value. Here's the model:
https://github.com/blackbaud/skyux-list-builder/blob/master/src/app/public/modules/list/state/search/search.model.ts

src/app/components/list/overview/index.html
Blackbaud-AlexKingman: > It looks like the last sentence there should be "This property accepts a ListFilterModel value" but you might want to check with a dev.

Yes an no. The last sentence should be this:

> This property accepts a function of type `(item: ListItemModel, filter: any) => boolean`

src/app/components/list/overview/index.html
Blackbaud-AlexKingman: Agree with @blackbaud-johnly , but have one thought -
Should it be "columns" instead of "types"? I'm not sure what "type" refers to in this context.

src/app/components/list/overview/index.html
Blackbaud-AlexKingman: This area gets super confusing for me too... I think this is an array of available sorts the end-user can apply. Not sure how to best put that in plain english. @Blackbaud-TrevorBurch am I reading this right?

src/app/components/list/overview/index.html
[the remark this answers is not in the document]
blackbaud-johnly: So if it accepts an array, it seems like we should definitely change "an item is" to "items are"

src/app/components/list/overview/index.html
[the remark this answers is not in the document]
blackbaud-johnly: Thanks! ... Let me know if anything else in here is unclear, Denise.

src/app/components/list/overview/index.html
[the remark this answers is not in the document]
blackbaud-johnly: Awesome. Thanks! ... I think that answers the only pending question, but let me know if anything else in here is unclear, Denise.

src/app/components/list/overview/index.html
blackbaud-johnly: Thanks, Alex. I should have asked what we meant by "types." I think subbing out "types" for "columns" in my suggested edit makes sense. ... Denise, let me know if I can do anything else here.

src/app/components/list/overview/index.html
blackbaud-johnly: If that's the case, what are "available sorts"? I definitely think it would be good to explain whatever this is in plain English, but I don't really follow what we are doing here. Based on https://github.com/blackbaud/skyux-list-builder/blob/master/src/app/public/modules/list/state/sort/label.model.ts, I'm wondering if a "sort" is a pre-filtered version of the list, but I'm not sure.

src/app/components/list/overview/index.html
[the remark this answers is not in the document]
Blackbaud-DeniseCisneros: Added link to `ListDataRequestModel` and added the `isToolbarDisabled` property

src/app/components/list/overview/index.html
Blackbaud-TrevorBurch: So this gets really tricky. That value really isn't used today. It appears to me that this is for `ListSortLabelModel` objects which are currently applied for the view being shown. This is for things such as when we have view switching and they can change the view from a grid to another view such as a repeater view (in the future). I'll be honest in that this only seems partially wired up when I dug into the code but in theory that is what this appears to be for. 

`global` is very similar but is those that are applied to all views. 

I'm honestly not sure the best way to put this as currently all sorts should show up as both global and available as I don't believe there is a current way to specify a non-global sort.

Blackbaud-AlexKingman: If we aren't 100% confident with these properties, then should we just leave it out?

blackbaud-johnly: I see the logic to leaving them out of the docs, but users could still find the properties and try to figure them out. If we have properties that aren't fully wired up or don't work, I think the responsible thing would be to be upfront about it. I think that for properties that are public and that we don't want users to use because they don't work, we should include a short boilerplate description that basically conveys that the properties are unfinished ("work in progress"?) and that we recommend against using them. What do you guys think?

Blackbaud-DeniseCisneros: I'll do whatever y'all tell me to and what you think is best, ha. Personally, I'm leaning towards John's suggestion of documenting `ListSortLabelModel` and `ListFieldSelectorModel` with a caveat saying we don't recommend using them. I can be swayed by whatever will make the most sense for the consumer though.

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
+  <sky-demo-page-properties sectionHeading="ListDataRequestModel properties">
+    <sky-demo-page-property
+      propertyName="filters"
+      isOptional="true"
+    >
+    Specifies a function to determine whether items are filtered. This property is required when using an in-memory data provider. This property accepts an array of <sky-code>ListFilterModel</sky-code> objects.
```
Blackbaud-AlexKingman: Can we link `ListFilterModel` its corresponding property table below?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+
+  <sky-demo-page-properties sectionHeading="ListFilterModel properties">
+    <sky-demo-page-property
+      propertyName="defaultValue"
+      isOptional="true"
+    >
+    Specifies the default value of the filter. When the filter is set to the default value, it is not applied. This property accepts <sky-code>any</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="dismissable"
```
Blackbaud-AlexKingman: The property name is `dismissible` (with an `i`).

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      isOptional="true"
+    >
+    Indicates whether users can dismiss the filter's summary item. This property accepts <sky-code>boolean</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="filterFunction"
+      isOptional="true"
+    >
+    Specifies a function to determine whether items are filtered. This property is required when using an in-memory data provider. This property accepts a function of type <sky-code>(item: ListItemModel, filter: any) => boolean</sky-code>.
```
Blackbaud-AlexKingman: > This property is required when using an in-memory data provider.

Same question as above. Do we need to outline this here? Is it really even true?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies the fields to sort. This property accepts an array of <sky-code>ListSortFieldSelectorModel</sky-code> objects.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="global"
+      isOptional="true"
+    >
+    Specifies the fields to sort. This property accepts an array of <sky-code>ListSortLabelModel</sky-code> objects.
```
Blackbaud-AlexKingman: Should we link `ListSortLabelModel` and `ListSortFieldSelectorModel ` to their respective tables above?

src/app/components/list/overview/index.html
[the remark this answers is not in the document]
Blackbaud-AlexKingman: @Blackbaud-TrevorBurch do you know if `ListDataRequestModel` is actually required?

src/app/components/list/overview/index.html
```diff
@@ -17,7 +17,7 @@
       propertyName="appliedFilters"
       isOptional="true"
     >
-      Specifies a set of filters to apply to list data. These filters also create a filter summary if <a routerLink="../filter">the <sky-code>sky-list-filter-summary</sky-code> component</a> is included. Acceptable values: <sky-code>Array of ListFilterModel</sky-code>.
+      Specifies a set of filters to apply to list data. These filters create a filter summary when the list includes <a routerLink="../filter">the <sky-code>sky-list-filter-summary</sky-code> component</a>. This property accepts an <a stacheRouterLink="." fragment="listfiltermodel-properties">array <sky-code>ListFilterModel</sky-code> properties</a>.
```
blackbaud-johnly: This link is broken. "filter" should be "filters"

src/app/components/list/overview/index.html
```diff
@@ -32,7 +32,7 @@
...
+    Specifies a data provider to obtain the data to display. The list component requires this property or the <sky-code>data</sky-code> property. Acceptable values: <sky-code>extends ListDataProvider</sky-code>. To extend ListDataProvider, provide a <sky-code>get</sky-code> function that accepts a <sky-code>ListDataRequestModel</sky-code> as an argument and returns an <sky-code>Observable ListDataResponseModel </sky-code>, and a <sky-code>count</sky-code> function that returns an <sky-code>Observable number</sky-code> that indicates how many total records are in the list. For lists that use <sky-code>dataProvider</sky-code> instead of <sky-code>data</sky-code>, consumers are responsible for managing <a stacheRouterLink="." fragment="listdatarequestmodel-properties">all <sky-code>ListDataRequestModel</sky-code> properties</a>. For examples of custom searching, filters, and paging, see the <a stacheRouterLink="." fragment="code">demo code</a> below.
```
blackbaud-johnly: Superminor, but can you please delete the extra space at the end of the `Observable ListDataResponseModel` code snippet?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      isOptional="true"
+    >
+    Specifies a function to determine whether items are filtered. This property is required when using an in-memory data provider. This property accepts an array of <a stacheRouterLink="." fragment="listfiltermodel-properties"><sky-code>ListFilterModel</sky-code> objects</a>.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="isToolbarDisabled"
+      defaultValue="false"
+      isOptional="true"
+    >
+    Specifies whether the search bar and filter button are disabled. This property accepts <sky-code>boolean</sky-code> values.
```
blackbaud-johnly: I'd change "Specifies" to "Indicates" because that's the tsandard wording for boolean properties. ... Also, I'd flip this around to use active voice: "Indicates whether to disable the search bar and filter button."

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    Specifies a set of fields to sort by. This property accepts <a stacheRouterLink="." fragment="listsortmodel-properties"> <sky-code>ListSortModel</sky-code> value</a>.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListFilterModel properties">
+    <sky-demo-page-property
+      propertyName="defaultValue"
+      isOptional="true"
+    >
+    Specifies the default value of the filter. When the filter is set to the default value, it is not applied. This property accepts <sky-code>any</sky-code> values.
```
blackbaud-johnly: I missed it the first time around, but the second sentence doesn't seem clear to me. What do we mean by "it is not applied"? ... Would it be clearer to flesh this out to something like " When users keep the default value, the filter does not effect the list."?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      propertyName="filterFunction"
+      isOptional="true"
+    >
+    Specifies a function to determine whether items are filtered. This property is required when using an in-memory data provider. This property accepts a function of type <sky-code>(item: ListItemModel, filter: any) => boolean</sky-code>.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="label"
+      isOptional="true"
+    >
+    Specifies a label for the filter's summary item. The label defaults to the <sky-code>value</sky-code> of the filter. This property accepts <sky-code>string</sky-code> values.
```
blackbaud-johnly: Is it just me, or is it kinda weird that the default label is the `value`? It seems kinda odd since the `value` can change.

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    Specifies the current value of the filter. This property accepts <sky-code>any</sky-code> values.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSearchModel properties">
+    <sky-demo-page-property
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies the columns to search. The columns correspond to <sky-code>field</sky-code> values that you specify with the <sky-code>data</sky-code> property. This property accepts an array of <sky-code>string</sky-code> values.
```
blackbaud-johnly: Do you think we should specify where the `data` property comes from? Either by calling it out (i.e., "...with the list component's `data` property") or linking to that section (i.e., making "the `data` property" a link to the section above)?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies the columns to search. The columns correspond to <sky-code>field</sky-code> values that you specify with the <sky-code>data</sky-code> property. This property accepts an array of <sky-code>string</sky-code> values. 
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="functions"
+      isOptional="true"
+    >
+    Specifies an array of functions that returns a <sky-code>boolean</sky-code> value of <sky-code>true</sky-code> when the search is successful. Acceptable values: <sky-code>(data: any, searchText: string) => boolean></sky-code>.
```
blackbaud-johnly: I'd adopt your wording from above and change "Acceptable values" to "This property accepts a function of type"

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    <sky-demo-page-property
+      propertyName="searchText"
+      isOptional="true"
+    >
+    Specifies the text to search. This property accepts <sky-code>string</sky-code> values.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSortFieldSelectorModel properties">
+    <sky-alert alertType="warning">The following properties are a work in progress and not recommended for use. </sky-alert>
```
blackbaud-johnly: Did you try this as a regular paragraph? It seems like a regular paragraph might be sufficient.

Also, what do you think of tweaking like this: "These properties are a work in progress, and we do not recommend using them."

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSortLabelModel properties">
+    <sky-alert alertType="warning">The following properties are a work in progress and not recommended for use.</sky-alert>
+    <sky-demo-page-property
+      propertyName="fieldType"
+      isOptional="true"
+    >
+    Specifies the label type. This property accepts <sky-code>string</sky-code>values.
```
blackbaud-johnly: We're missing a space between the code snippet and "values."

Also, it might not matter since we recommend against using these properties, but I don't know what "label type" is. That doesn't really mean anything to me.

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      isOptional="true"
+    >
+    Specifies the fields to sort. This property accepts <sky-code>string</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="global"
+      isOptional="true"
+      defaultValue="false"
+    >
+    Specifies all fields to sort. This property accepts <sky-code>boolean</sky-code> values.
```
blackbaud-johnly: The first sentence doesn't match up with this being a boolean property. I assume the description should be something more like "Indicates whether to sort all fields." Or something like that.

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      isOptional="true"
+      defaultValue="false"
+    >
+    Indicates whether to sort in descending order. This property accepts <sky-code>boolean</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="text"
+      isOptional="true"
+    >
+    Specifies the text for the label. This property accepts <sky-code>string</sky-code> values.
```
blackbaud-johnly: What is the label for?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+  <sky-demo-page-properties sectionHeading="ListSortModel properties">
```
blackbaud-johnly: Should this section have that warning too?

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+    <sky-demo-page-property
+      propertyName="available"
+      isOptional="true"
+    >
+    Specifies the list of available views to sort. This property accepts an array of <sky-code> ListSortLabelModel</sky-code> objects.
```
blackbaud-johnly: I don't want to keep digging, but do we need a table for the `ListSortLabelModel` properties?

Also we have an extra space before `ListSortLabelModel`.

src/app/components/list/overview/index.html
```diff
@@ -75,6 +75,180 @@
...
+      propertyName="fieldSelectors"
+      isOptional="true"
+    >
+    Specifies the fields to sort. This property accepts an array of <a stacheRouterLink="." fragment="listsortfieldselectormodel-properties"><sky-code>ListSortFieldSelectorModel</sky-code> objects</a>.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="global"
+      isOptional="true"
+    >
+    Specifies the fields to sort. This property accepts an array of <a stacheRouterLink="." fragment="listsortlabelmodel-properties"><sky-code>ListSortLabelModel</sky-code> objects</a>.
```
blackbaud-johnly: The first sentence here is the same as the first sentence for the previous property. That doesn't seem right.

src/app/components/list/overview/index.html
blackbaud-johnly: I'm going to tweak this slightly...

src/app/components/list/overview/index.html
blackbaud-johnly: Guessing that either my question didn't make sense or didn't require any change...

src/app/components/list/overview/index.html
blackbaud-johnly: Updating this with the first suggested tweak. Feel free to undo this change if you deliberately left this as is.

src/app/components/list/overview/index.html
blackbaud-johnly: Assuming we decided this doesn't matter since we recommend against using the property...

src/app/components/list/overview/index.html
blackbaud-johnly: Assuming that we decided explaining label type doesn't matter since we recommend against using this property...

src/app/components/list/overview/index.html
blackbaud-johnly: Just following up: What I meant was does this section require the sentence between the header and the table that explains "These properties are a work in progress, and we do not recommend using them."?

src/app/components/list/overview/index.html
blackbaud-johnly: Is it accurate that the `global` property and the `fieldSelectors` property both specify the fields to sort?

src/app/components/list/overview/index.html
blackbaud-johnly: @Blackbaud-DeniseCisneros, can we resolve this? Or is anything still pending?

src/app/components/list/overview/index.html
blackbaud-johnly: @Blackbaud-DeniseCisneros, can we resolve this? Or is anything still pending?

PR #696, opened 2019-10-11
Added Window service docs

Blackbaud-MatthewMiles:
Added SkyAppWindowRef docs to surface this functionality from skyux/core.

This is for Issue https://github.com/blackbaud/skyux2-docs/issues/266

PR #703, opened 2019-10-16
Document Window service

blackbaud-johnly:
Replaces https://github.com/blackbaud/skyux2-docs/pull/696 so that checks will pass.


blackbaud-johnly:
FYI @Blackbaud-MatthewMiles that this is getting in. Thanks for the docs contribution!

PR #710, opened 2019-10-20
Tree view edits

blackbaud-johnly:


src/app/components/tree-view/index.html
```diff
@@ -86,7 +90,7 @@ <h3>Split view</h3>
       <h3>Expand/collapse nodes</h3>
         <sky-docs-visual-example>
           <sky-docs-visual-example-text>
-            <p>Clicking the chevron button expands or collapses the node.</p>
+            <p>The expand/collapse button expands and collapses the node.</p>
```
blackbaud-johnly: I changed this back to `expand/collapse button` because that's what we call it in the Anatomy section. If we want to use `chevron button` instead, I think we should adjust all instances of the terminology.

src/app/components/demo-components.service.ts
```diff
@@ -1161,7 +1161,7 @@ export class SkyDemoComponentsService {
       {
         name: 'Tree view',
```
blackbaud-johnly: Is “tree view” the correct name for the module and the correct terminology to talk about it? I left it as-is, but we don’t use “view” in the component names or in the naming patterns in the GitHub repo, and it doesn’t seem like the library that we are building off of uses “view” either. Is this really the correct way to refer to this module? And if so, why isn’t this wording convention built into the names of the components and such? (The Development tab had multiple instance of just “tree” instead of “tree view” and I updated them for consistency, but I’m not sure this change that was really the right call.)

src/app/components/tree-view/index.html
```diff
@@ -5,17 +5,21 @@
...
-      This component is a release candidate. We do not plan to make any breaking changes the component before the official release, but such changes are possible.
-    </sky-alert>
+    <stache-include fileName="components/release-candidate-warning.html"></stache-include>
   </sky-demo-page-summary>
   
   <sky-demo-page-example>
     <sky-angular-tree-view-demo></sky-angular-tree-view-demo>
   </sky-demo-page-example>
 
+	<div class="sky-margin-stacked-content-section">
```
blackbaud-johnly: What’s the deal with reordering items in the list? Messing around with the demo, I’ve been able to drag and drop list items, but it’s extremely temperamental and I can’t do it consistently. I don’t see a setting for this, so I don’t think it’s a property on our components. So does that mean it’s a feature of our components that is enabled by default? If so, should we have something about it in the design docs? Or is this inherited by the Angular components that we are building on? Also, if the feature is this fluky, should we try to disable it by default? Just clicking around, I was able to accidentally drag a cat to the dogs section in the demo, but I wasn’t able to drag it back to the correct section.

src/app/components/tree-view/index.html
```diff
@@ -49,23 +53,23 @@ <h3 class="sky-text-success">Use when</h3>
...
+            <p>You can use the single-select and multiselect modes to allow users to select list items.</p>
           </sky-docs-visual-example-text>
           <sky-docs-visual-example-image>
             <img src="~/assets/img/guidelines/angular-tree-view/option-select.png" />
           </sky-docs-visual-example-image>
         </sky-docs-visual-example>
         <sky-docs-visual-example>
           <sky-docs-visual-example-text>
-            <p>With cascading selection, when users select a node the descendant nodes are also selected. Use this mode if it's likely that users will want to select a node and all its descendants.</p>
+            <p>You can use the cascading selection mode to select descendant nodes when users select items. Use this mode when users are likely to select all descendants within nodes.</p>
```
blackbaud-johnly: Should cascading selections really be listed as an option in our design docs? As far as I can tell, it’s not an option that is set through our components, so should we really call it out as an option when we don’t call out the other options for the components that we are building on top of? Same question for the way we highlight it in the demo.

src/app/components/tree-view/index.html
```diff
@@ -49,23 +53,23 @@ <h3 class="sky-text-success">Use when</h3>
       <h3>Selection</h3>
```
blackbaud-johnly: Should we call out read-only mode and toolbar somewhere in the Options section of the Design docs? It seems like we should since these are valid options of the components we built. I especially think this is the case for the read-only mode because we call it out in the Use when section and then in the Options section, we start right in on the selection modes. And we highlight both items in the demo.

src/app/components/tree-view/index.html
```diff
@@ -74,7 +78,7 @@ <h3>Selection</h3>
       <h3>Split view</h3>
```
blackbaud-johnly: Just my 2 cents, but the split view example doesn’t seem like an option to me. It seems like a use case that should be in the Use when section, not an option for this section. I mean, this isn't an option you select on the tree view components. It's an implementation of the tree view components inside of another component.

src/app/components/tree-view/index.html
```diff
@@ -39,7 +43,7 @@ <h3 class="sky-text-success">Use when</h3>
     <sky-demo-page-content sectionHeading="Anatomy">
       <sky-docs-anatomy imagePath="~/assets/img/guidelines/angular-tree-view/anatomy.png">
         <sky-docs-anatomy-item anatomyNumber="1">Expand/collapse button</sky-docs-anatomy-item>
-        <sky-docs-anatomy-item anatomyNumber="2">List item content</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="2">List item</sky-docs-anatomy-item>
         <sky-docs-anatomy-item anatomyNumber="3" isOptional="true">Checkbox</sky-docs-anatomy-item>
         <sky-docs-anatomy-item anatomyNumber="4" isOptional="true">Context menu</sky-docs-anatomy-item>
```
blackbaud-johnly: Should we really call out context menus here? And if so, should we include an entry in the Options section? We highlight context menus in the demo, but as far as I can tell, this is not an option of our components and is something that consumers need to set on the Angular component we are building off of. So it seems like maybe this shouldn't really be here. But if we list cascading as an anatomy item and in the demo, then shouldn’t we also call out this option in the Options section?

src/app/components/tree-view/index.html
```diff
@@ -103,18 +107,14 @@ <h3>States</h3>
     <sky-demo-page-content sectionHeading="Related information">
       <h3>Components</h3>
       <ul>
-        <li><a routerLink="/components/flyout">Split view</a></li>
+        <li><a routerLink="/components/split-view">Split view</a></li>
```
blackbaud-johnly: Should we include the list component in the related components? I mean, if the hierarchical view isn't necessary, that's the first alternative, isn't it?

src/app/components/tree-view/index.html
```diff
@@ -126,20 +126,20 @@ <h3>Components</h3>
       sectionHeading="Angular tree view components"
     >
       <p>
-        To apply the SKY UX themes and additional functionality to the <sky-code>angular-tree-component</sky-code>, you must include the following components.
+        <sky-code>SkyAngularTreeModule</sky-code> provides SKY UX components and styles to complement the Angular tree component and apply SKY UX themes and functionality to hierarchical list views. For information about the Angular tree component and its capabilities, see the <a href="https://angular2-tree.readme.io/docs">Angular tree component documentation</a>.
```
blackbaud-johnly: Is “angular-tree-component” the actual selector for the Angular component? I can’t tell from there docs, but that doesn’t seem right. I see in some of there reference docs that they have it presented this way (but not as a code snippet), but I see other places where the refer to it generically. I changed all of our references like this with the code snippet formatting to the generic “Angular tree component” with no code snippet formatting because it didn't seem right. But since I’m not sure if that’s correct, I just wanted to mention that I made this change throughout..

src/app/components/tree-view/index.html
```diff
@@ -126,20 +126,20 @@ <h3>Components</h3>
...
       </p>
 
       <sky-demo-page-property
         propertyName="sky-angular-tree-wrapper"
```
blackbaud-johnly: How do users control whether to display the context menu? We have a component in the module, but we don’t document it in the Development tab. I think Alex said the toolbar component is for internal use only, but I don’t remember if that’s also the case for context menu. If not, I think we need to document that component. But if so, how do consumers of this module determine whether to display the context menu? It seems like we need something about this somewhere. If not, should we remove it as an option in the demo and Design docs?

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: I didn't think we were supporting drag & drop in this initial release, @Blackbaud-AlexKingman is this something built into the library that we need to disable?

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: For the UX pattern, context menus are an option that the designer needs to consider. If we don't include them in the example, designers will think they can't use them. Need to think about whether to include it in the Options section.

I am not sure how the consumer technically includes them.

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: Yeah I will add something, probably just a text description since I don't know how to illustrate things that aren't allowed ;).

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: I think the issue is the header saying "Split view" instead of "Active mode" which is intended for use specifically with split views.

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: leaving it as expand/collapse button seems ok

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: probably should link to Grid and Repeater as generic views.

src/app/components/tree-view/index.html
blackbaud-johnly: Sounds good.

src/app/components/tree-view/index.html
blackbaud-johnly: Maybe I'm just confused here, but this still doesn't seem like an option to me. I mean, isn't this just default behavior? If the tree view is included in a split view, then isn't "active mode" just the the indicator that the component implements automatically when users select an item because the tree view isn't in a select mode? I may be splitting hairs here, but I feel like I'm missing something.

src/app/components/tree-view/index.html
blackbaud-johnly: Makes sense. Thanks!

src/app/components/tree-view/index.html
blackbaud-johnly: Sounds good. I can add links to grid and repeater. But what about list?

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: The issue is that the 3rd party library has wonky states/modes. From a UX pattern perspective, "Active mode" is a thing regardless of how it's implemented.

blackbaud-johnly: OK. So should we change the header to "Active mode" and update the text to spell out what active mode is and how it pertains to split views?

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: not sure, since tree view doesn't work with List as far as I know

src/app/components/tree-view/index.html
Blackbaud-ToddRoberts: yeah

src/app/components/tree-view/index.html
blackbaud-johnly: I added links to grid and repeater, but not to list.

src/app/components/tree-view/index.html
blackbaud-johnly: I changed the header to "Active state" and fleshed the paragraph out a bit to try to emphasize that as the option.

src/app/components/demo-components.service.ts
Blackbaud-AlexKingman: Yes, we have a consistency problem here. Our pattern for 3rd party library components is to use the library's name. The only thing I don't care for in this particular scenario is that the third party library has the words "Angular" and "component" in its name. Bleh. With that said, I think we should probably still follow the pattern to avoid confusion. After all, this is our SKY-specific implementation of `angular-tree-component`. Here's some examples of how I think we should refer to it:

The npm package:
`@skyux/angular-tree-component`

The repo:
`skyux-angular-tree-component`

The file directory:
`/angular-tree-component/...`

How we *should* refer to the component:
`Angular tree component`

What do you think?

blackbaud-johnly: It makes sense to me to follow the pattern of using the third-party library's name, and I'm good with all these naming changes. But I think we should confirm this with @Blackbaud-ToddRoberts first to make sure he's OK with it.

src/app/components/tree-view/index.html
blackbaud-johnly: I'm going to resolve this comment based on Alex's comments above on the name of the module.

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: We haven't spent any time supporting it for this release. I could try and see if there's a way to prevent it, but I'd have to mess with the native 3rd-party library. The demo *shouldn't* be draggable, though. @blackbaud-johnly How did you get drag/drop enabled?

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: You are correct - its not something we had to build, as it was already supported by the 3rd-party library. With that said, I think its important to still talk about best practices in design, even if its not a particular function of our component. These are scenarios that our consumers will come across and have question about. That's just my 2 cents.

src/app/components/tree-view/index.html
blackbaud-johnly: I didn't do anything to enable drag/drop. I just noticed when I was clicking around selecting/deselecting items that I could click and drag. It's tempermental, but you can do it. I just tested it again, and was able to drag a couple cats to the dogs section. It's hit and miss on whether I can get items to be draggable though.
image

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: That's probably for the better. You're right - there is no `angular-tree-component` selector. I should have referred to the generic component like you did above. 👍

src/app/components/demo-components.service.ts
Blackbaud-ToddRoberts: IMO the info architecture/nav should be implementation-agnostic both for ease of discovery (users would be looking for "Tree view" not "Angular tree component") and future-proofing if we ever changed implementation or added additional implementations (e.g. native mobile components)

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: Yes, we could use some clarity here. There's no property or setting for a consumer to use context menus. They simply just include the dropdown component inside the `sky-angular-tree-node`. We've added built-in support for the dropdown, such that we'll make sure it gets styled and positioned correctly in the node. If the consumer adds it to the template, we do the rest.

What's the best way to describe that?

src/app/components/demo-components.service.ts
blackbaud-johnly: So then for the examples Alex provides above, should we go with:

The npm package:
`@skyux/tree-view`

The repo:
`skyux-tree-view`

The file directory:
`/tree-view/..`.

How we should refer to the component:
`tree view component`

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: Whoa. You definitely should not be able to do that. I'll look into it.

What browser/OS are you using?

src/app/components/demo-components.service.ts
Blackbaud-ToddRoberts: Repo and npm package I am not worried about, it's probably valuable to be implementation-specific there. It's more in the docs navigation and pattern-related (i.e. non-implementation-related) content that we should be agnostic.

Blackbaud-AlexKingman: Todd makes a good point about how we should lead with the UX pattern, and not the implementation. That makes me think `tree view` would be better for the navigation of the docs site and the title of the page. However, the repo, npm package, and component should still be called `angular-tree-component` because those ARE implementation-specific. We wouldn't add a "react" component to the same package. It would just be a new package - for example, `@skyux/react-super-treeview` (an actual react open-source library).

Blackbaud-AlexKingman: FWIW - we'll have this same problem with the ag-Grid pr: https://github.com/blackbaud/skyux2-docs/pull/711

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: After @Blackbaud-ToddRoberts 's comment above how this page is more about the "Tree view" pattern in general, and NOT the specific implementation, I think it further supports keeping UX guidelines like this in.

src/app/components/tree-view/index.html
blackbaud-johnly: Chrome on Windows 10.

src/app/components/tree-view/index.html
blackbaud-johnly: Works for me. Just wanted to make sure. I'll resolve this conversation.

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: Reproduced. Filed a bug: https://github.com/blackbaud/skyux-angular-tree-component/issues/6

I'll get on it, thanks!

src/app/components/tree-view/index.html
blackbaud-johnly: It seems like we should include this info in the Design docs. We have a section there about Active state that is tied to the split view component, and it seems like this. What do you guys think of adding a section under *Options* along the lines of:

*Context menu*
When users need to perform actions on items in hierarchical lists, you can place dropdown components in the tree node components for those items. The tree node component automatically handles handles styling and positioning for the context menu.

src/app/components/demo-components.service.ts
blackbaud-johnly: OK. Based on this decision, I think most of the docs are OK as is, but I'll take another pass to make sure.

src/app/components/tree-view/index.html
Blackbaud-AlexKingman: Love it. :shipit:

blackbaud-johnly: Good deal. I'll make this update and double-check the naming tonight. ... @Blackbaud-ToddRoberts, let me know if you have any edits here.

Blackbaud-ToddRoberts: @Blackbaud-AdamFunderburk brought up that we should be sure the code example makes it clear how to do this

src/app/components/demo-components.service.ts
blackbaud-johnly: I went through, and the docs seem OK. ... For the naming patterns, should "angular" be part of the file names and folder paths for the demos?

src/app/components/tree-view/index.html
blackbaud-johnly: Resolving since this is a bug with the component and not a docs issue.

src/app/components/tree-view/index.html
blackbaud-johnly: Links seem to work, so resolving this issue.

src/app/components/tree-view/index.html
blackbaud-johnly: I added a new section with the text in this conversation and linked to the dropdown component docs. ... The demo includes an option to turn context menus off and on, and I see a dropdown component in the code samples. @Blackbaud-AlexKingman, do you think what we have is sufficient to address Adam's point?

Blackbaud-AlexKingman: Yep, I think we're covered here.

src/app/components/demo-components.service.ts
blackbaud-johnly: @Blackbaud-AlexKingman: Any thoughts on the file names and folders for the demos? Should they include "angular"?

Blackbaud-AlexKingman: Hmm... the folder names are exposed to the end user, so I feel like they should match our descriptors on the docs page. If we refer to this page as `Tree view`, then the folder should probably be `./tree-view` as well.

Following this pattern, you can make an argument that our demo is a specific implementation. So perhaps the demo files should match the 3rd-party library name. For example:
```
./angular-tree-component/angular-tree-component-demo.component.html
./angular-tree-component/angular-tree-component-demo.component.ts
./angular-tree-component/angular-tree-component-demo.module
```

What do you think?

Blackbaud-ToddRoberts: demo files matching 3rd party library name seems correct to me.

src/app/components/tree-view/index.html
blackbaud-johnly: @Blackbaud-ToddRoberts, are you good with no image?

src/app/components/demo-components.service.ts
blackbaud-johnly: I tried to update the demo filenames but ran into errors. I spent some time troubleshooting but just got different errors. Rather than spin my wheels troubleshooting, Alex and I decided that he would just make those changes in his main docs branch. I'm resolving this conversation since we'll address the filenames there.

src/app/components/tree-view/index.html
blackbaud-johnly: Resolving this thread because Todd took care of the updates to the Design tab, and I added context menu info to the Development tab.

Issue #714, opened 2019-10-23
Document and demo repeater reorder functionality


Blackbaud-TrevorBurch:
We need to document and demo the reorder functionality for repeaters that was added in `3.5.0`

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/724/.

PR #724, opened 2019-10-30
Document reorder property for repeater items

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/714. Also documents missing public event `isSelectedChange`.


src/app/components/repeater/index.html
```diff
@@ -61,6 +61,13 @@
     >
     Indicates whether the repeater item's checkbox is selected. When users select the repeater item, the specified property on your model is updated accordingly. This property accepts <sky-code>boolean</sky-code> values.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="reorderable"
```
Blackbaud-AlexKingman: `reorderable` is both an input property for the repeater component and the repeater item component. Here, we've only documented it on the repeater item component, but all our demos/tests show it only used on the parent repeater component. This should probably be moved to a repeater component property.
@Blackbaud-TrevorBurch might need to confirm this.

blackbaud-johnly: Thanks for catching that. I dug into the source file for the repeater item component because that's what I noticed on the PR. I'll check the source file for the repeater component as well. Thanks!

blackbaud-johnly: I updated the docs to reflect the second property. Are the descriptions corrrect?

PR #731, opened 2019-11-01
Made blog posts anchors

Blackbaud-BobbyEarl:


Blackbaud-DeniseCisneros:
If you're going to use anchors, should we turn on the TOC so someone could easily navigate to the anchors?

Blackbaud-DeniseCisneros:
One more comment 😬 The TOC is a little noisy with the "Design Update - Date" format. If the blog posts are mostly design updates, does it make sense to leave that off and only list the type of update when it's something other than a design update?

Blackbaud-BobbyEarl:
These are all great questions @Blackbaud-DeniseCisneros.  I believe the true crux the problem is Stache isn't (currently) in an ideal state when it comes to blogs.  There was a huge effort places in the first version of Stache but it hasn't really been a priority in the current version.

My main concern was making the headers linkable.  I do think there's plenty of opportunity to continue tweaking this page and layout, but I'm going to push that off in to a future PR.

PR #733, opened 2019-11-04
Progress indicator edits

blackbaud-johnly:


src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+    
+    <sky-tab
+      tabHeading="Design">
+    
+      <sky-demo-page-content sectionHeading="Usage">
+        <h3 class="sky-text-success">Use when</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>Use waterfall progress indicators for long or complex setup processes that users typically complete only once. Waterfall indicators are ideal when events outside of the system force users to complete tasks in multiple sittings or when users must enter large amounts of data.</p>
+            <p>Use waterfall progress indicators when processes include at least three steps and the number and order of steps do not change.</p>
```
blackbaud-johnly: We say the best experience requires at least three steps. Should we also recommend an upper limit for the best experience?

blackbaud-johnly: Also, should this second paragraph really be treated as part of the first Use when item? It seems like it’s completely distinct from the previous paragraph to me. And it even feels like maybe it should be restated as a Don’t use when item. I think that what we are trying to say is to NOT use waterfalls when the number or order of steps changes, not TO use it whenever processes have stable steps. Same with the number of steps. It seems like maybe we want to recommend NOT to use this pattern for fewer than 3 steps, not to recommend it for process with more than 3 steps.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div>
+              <img src="~/assets/img/guidelines/progress-indicator/waterfall-progress-usage-1.png" />
```
blackbaud-johnly: “Setup” should be “Set up” because it’s a verb, not a noun or adjective. I have a PR to fix this in the demo (although it's on hold while I work out how to split out the demos). Should we fix this in this image and the one below as well?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+              <div class="caption caption-success sky-text-success">Do use waterfall progress indicators for complex configuration processes that users may complete in stages.</div>
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>Use waterfall progress indicators when the complexity of steps varies drastically. For example, one step may require a full-screen modal for complex interactions, while another step requires a small modal for a few simple inputs. Waterfall steps can open <a routerLink="/components/wizard">wizards</a> if necessary.</p>
```
blackbaud-johnly: I went to the modal docs to see if “full-screen modal” was the correct terminology. In those docs, we use “full-screen” and “full-page” interchangeably. I think we should we stick to one or the other. Do you know which is preferred?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+          </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div class="sky-margin-stacked-separate">
+              <img src="~/assets/img/guidelines/progress-indicator/waterfall-progress-usage-2.png" />
```
blackbaud-johnly: Why don’t the links in steps 1 and 2 and the button in step 3 line up? In the previous image and the anatomy example, the buttons line up with the links. However, the links and buttons don’t line up in the demo. It seems like the spacing should be consistent. Do we have a problem with inconsistent spacing in the way the component works? Or maybe a problem with the Design images? … More or less the same question about the second image of the second Use when item and the link in step 1 and the button in step 3 line up. And same question for the link and button in the first image of the first Don’t use when item. And in the first image of the third Don’t use when item, we have 4 links, and only 2 line up. That doesn’t seem right.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+              <div class="caption caption-success sky-text-success">Do use modals of the appropriate size for the interactions in each waterfall step.</div>
+            </div>
+            <div>
+              <img src="~/assets/img/guidelines/progress-indicator/waterfall-progress-usage-3.png" />
+              <div class="caption caption-danger sky-text-danger">Don't put inputs inside waterfall progress indicators.</div>
```
blackbaud-johnly: Should this be a separate Don’t use when item below instead of a secondary image/caption here? I can definitely see where it’s tangential to the Use when info, but it seems to me like it’s distinct enough to warrant it’s own Don’t use when item below. (Either that or a separate Use when item here that says to use waterfall progress indicators when the actions that users need to perform are going to be performed in inputs on modals that are opened from the waterfall indicator but not by inputs directly on the waterfall indicator.)

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+    
+        <h3 class="sky-text-danger">Don't use when</h3>
```
blackbaud-johnly: Should there be a Don’t use when item calling out not to use the waterfall for processes that users only monitor? And a link to passive indicators? Not sure if this is necessary, but it seemed worth asking.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+        </sky-docs-visual-example>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Anatomy">
+        <sky-docs-anatomy imagePath="~/assets/img/guidelines/progress-indicator/waterfall-progress-anatomy.png">
+          <sky-docs-anatomy-item anatomyNumber="1">Step indicator</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="2">Step label</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="3">Completed step action</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="4">Active step action</sky-docs-anatomy-item>
+          <sky-docs-anatomy-item anatomyNumber="5">Waterfall title</sky-docs-anatomy-item>
```
blackbaud-johnly: Why is the waterfall title #5? Given that it’s at the top of the page, it seems like it should be #1. But given that it’s included in the Options section, I assume that the problem is that this should actually be marked as optional. Is that correct?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+        
+        <h3>Start over action</h3>
+        <div class="sky-text-block">
+          <p>You can include a start over action to let users erase all saved data from the waterfall process. This returns user to the first step or to an earlier point in the workflow if decisions occured before the waterfall progress indicator.</p>
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Behavior & states">
+        <div class="sky-text-block">
+          <p>Waterfall progress indicators guide users through one active step at a time. When users complete a step, the progress indicator updates its step indicator and step action to the completed state and updates the next step indicator and step action to the active state. Users can review and edit completed steps, but this does not undo the completed state.</p>
```
blackbaud-johnly: If we want to show examples like Todd suggested in the other PR for this, I think we could use an image of side-by-side progress indicators where the current step advances from left to right to show the transition from one step to the next. But I don't know how useful this would be.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+          <p>The final step summarizes the changes and requests user confirmation. You do not need to include all data that users enter in the summary, but confirmation is required. After users complete the final step, replace the waterfall progress indicator with a page that displays the finished or active state of the system. Any page type is allowed, including record pages, work centers, and list pages.</p>
```
blackbaud-johnly: If we want to show examples like Todd suggested, I think we could also use side-by-side images here with a progress indicator on the last step to the left and finished state image on the right. This seems to me like it might be useful.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Content">
```
blackbaud-johnly: I tightened up the paragraphs in the Content section, and one change was to remove the examples in the parentheses in the first paragraph because they were incomplete. None of the examples followed the <verb> <direct object> format that we recommend in that paragraph because they are all missing the direct objects. So I deleted the examples to avoid the confusion of the guideline not matching the examples. If you think it’s important to have examples of step labels, we can add them back, but I would recommend including direct objects in the examples.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+        <div class="sky-text-block">
+          <p>Step labels in waterfall progress indicators start with a number and concisely describe the objective. Use the <b>&lt;verb&gt; &lt;direct object&gt;</b> format. In the label for the last step, describe the cumulative change that occurs at the end of the process.</p>
+          <p>Step action labels closely follow step labels but do not need to be identical. For the first step action label, be generic and conversational (“Get started”), and for completed step actions, use verbs appropriate to post hoc actions (“Change,” “Edit”).</p>
```
blackbaud-johnly: What is the preferred format for action labels? Should they also follow the <verb> <direct object> format that we recommend for the step labels? Users could infer this pattern from the previous paragraph or assume that because we don’t state it explicitly for actions, it doesn’t apply here. I’d remove the ambiguity by being explicit. If this rule applies to both step labels and step actions, I’d make that clear in the first paragraph and then either update or remove the examples for the action labels since they don’t include direct objects. But if it doesn't apply here, I'd explicitly state the recommended format in this paragraph.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Layout">
+        <div class="sky-text-block">
+          <p>Waterfall progress indicators are 400 pixels wide and centered inside the page. The waterfall title is the page title. Do not combine waterfall indicators with other content or actions.</p>
```
blackbaud-johnly: Should the second sentence about how the waterfall title is the page title be in the Content section?

src/app/components/progress-indicator/passive-progress-indicator/index.html
```diff
@@ -0,0 +1,276 @@
...
+            </div>
+            <div>
+              <img src="~/assets/img/guidelines/progress-indicator/passive-progress-usage-5.png" />
+              <div class="caption caption-success sky-text-success">Do roll up steps into a single status in lists. Use a container such as <a routerLink="/components/flyout">flyouts</a> or <a routerLink="/components/popover">popovers</a> to display details from passive progress indicators.</div>
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Anatomy">
```
blackbaud-johnly: Should the Anatomy section have an entry to call out the pending step? It is formatted differently than the regular step label. It seems like we should have a completed step label entry and a pending step entry since they seem to be as distinct as the completed step action and the active step action in the waterfall indicator anatomy.

src/app/components/progress-indicator/passive-progress-indicator/index.html
```diff
@@ -0,0 +1,276 @@
...
+              <img src="~/assets/img/guidelines/progress-indicator/passive-progress-usage-1.png" />
+              <div class="caption caption-success sky-text-success">Do use passive progress indicators to track sequential steps that other users take.</div>
+            </div>
+          </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+    
+        <h3 class="sky-text-danger">Don't use when</h3>
+        <sky-docs-visual-example>
+          <sky-docs-visual-example-text>
+            <p>Don't use passive progress indicators for tasks that require user interaction.</p>
```
blackbaud-johnly: Should we say to use waterfall indicators or wizards instead and link to those topics?

src/app/components/progress-indicator/passive-progress-indicator/index.html
```diff
@@ -0,0 +1,276 @@
...
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Options">
+        <h3>Step details</h3>
+        <div class="sky-text-block">
+          <p>You can provide additional details about completed or current steps. Use step details to provide extra context to users, but avoid too many details or the implication that users should take action.</p>
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Content">
```
blackbaud-johnly: We call out to use past tense verbs for step labels, but in the examples we only do this for completed steps. For pending steps in the examples, we generally use present tense verbs (mostly gerunds, but not always). The Content section also calls out info about datestamp and timestamp without specifying that it is for completed steps. I’ve updated the info to call out that the guidance is for completed steps. Do we need to provide additional guidance for pending steps? Also, should we provide any guidance on the cases when not to include datestamps? Or on the cases when to include timestamps? I'm not sure that "in most cases" and "if necessary" are really all that clear. (We may not be able to provide reasons examples for when to deviate from the norm here, but if not, we could still possibly make the rationale more clear. For example we could change the date stamp info to "and include datestamps in parentheses unless you have a compelling reason not to." I'm not sure if that's any better though, so just throwing this out there for your consideration.)

src/app/components/progress-indicator/passive-progress-indicator/index.html
```diff
@@ -0,0 +1,276 @@
...
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Layout">
+        <div class="sky-text-block">
+          <p>Use passive progress indicators one at a time. When they apply to record views, use them in page summaries or other containers such as tiles.</p>
+          <p>To display passive progress indicators for items in lists, use dynamic containers such as <a routerLink="/components/flyout">flyouts</a> or <a routerLink="/components/popover">popovers</a>.</p>
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Related information">
```
blackbaud-johnly: Should we link to waterfall indicators or wizards?

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
```diff
@@ -0,0 +1,325 @@
...
+          <p>If waterfall indicators require extra context such as descriptions, links, or instructions, include it after the waterfall title and before the first step.</p>
+        </div>
+      </sky-demo-page-content>
+    
+      <sky-demo-page-content sectionHeading="Related information">
```
blackbaud-johnly: Should we link to passive indicators?

src/app/components/progress-indicator/passive-progress-indicator/index.html
Blackbaud-AdamFunderburk: Agreed.

src/app/components/progress-indicator/passive-progress-indicator/index.html
Blackbaud-AdamFunderburk: image

src/app/components/progress-indicator/passive-progress-indicator/index.html
Blackbaud-AdamFunderburk: Yup.

src/app/components/progress-indicator/passive-progress-indicator/index.html
Blackbaud-AdamFunderburk: Updated.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: Yup.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I think you're asking about the horizontal positions of the actions inside the waterfall.

The component _centers_ those actions to the page (well, to its <div> I think, but _that's_ centered to the page as well). There's different horizontal space between the step indicators/labels and the step actions, but that's intentional.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: I think "full-screen" is more precise.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
Blackbaud-AdamFunderburk: Good catch. Fixed 3 images.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Sounds like this is probably working as intended. It looked like some links and action buttons were lined up on their lefthand starting point (i.e., the first letters of the links and the left edge of buttons) while others weren't. But if the items are centered, then I guess it's just that some labels look like they line up because their text is of a similar length.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: I left "full-screen" as is here, and I just pushed updates to this branch for the modal docs to use it consistently there too.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: I still see "Setup" instead of "Set up" when I preview the PR, so I assume we still need to swap out the images. ... The demo update should be good to go, and I'll coordinate with Steve to get that in.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to https://github.com/blackbaud/skyux2-docs/pull/625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

src/app/components/progress-indicator/waterfall-progress-indicator/index.html
blackbaud-johnly: Ported over to #625.

PR #737, opened 2019-11-12
Fix list grid sticky settings link

jeffbdye:
Link seemed to 404 previously, I think this is where it was pointing now that all the integrations are on the skyux site.


blackbaud-johnly:
Thanks, @jeffbdye. It looks like in addition to the busted link, we need to make a couple other doc tweaks to get the description for this property in line with the changes we recently made to other property descriptions. So closing this in preference of https://github.com/blackbaud/skyux2-docs/pull/738.

PR #739, opened 2019-11-13
Remove popover docs

Blackbaud-SteveBrush:
Waiting on: https://blackbaud.visualstudio.com/Products/_git/skyux-host/pullrequest/80678?_a=overview


src/app/components/popover/index.html
```diff
@@ -1,283 +1,3 @@
...
-    </sky-demo-page-code>
-  </sky-demo-page-example>
-
-  </sky-tab>
-
-</sky-tabset>
-
-</sky-tabbed-demo-page>
+<sky-docs-url-redirect
+  url="https://developer.blackbaud.com/skyux-popovers/docs/popover"
```
Blackbaud-BobbyEarl: Since we'll eventually have 50 of these, should we factor out as much of the URL in to the component as possible?

Issue #695, opened 2019-10-11
Add `barebones` as another `--template` type for `skyux new`


Blackbaud-SteveBrush:
Here: https://developer.blackbaud.com/skyux/learn/reference/cli-commands/new#options

See https://github.com/blackbaud/skyux-sdk-template/issues/6

Blackbaud-DeniseCisneros:
Fixed and closed in https://github.com/blackbaud/skyux2-docs/pull/743

PR #743, opened 2019-11-13
Added barebones as template option

Blackbaud-DeniseCisneros:
References https://github.com/blackbaud/skyux2-docs/issues/695

### DIFF
--- a/src/stache/includes/cli/template.html
+++ b/src/stache/includes/cli/template.html
@@ -5,6 +5,9 @@
     <sky-code>-t addin</sky-code> — References the <a href="https://github.com/blackbaud/skyux-sdk-template-addin">SKY UX Add-in template</a> to create <a href="https://developer.blackbaud.com/skyapi/docs/addins">SKY Add-ins</a> that allow developers to build features that integrate with Blackbaud solutions.
   </li>
   <li>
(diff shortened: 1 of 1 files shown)


src/stache/includes/cli/template.html
```diff
@@ -4,6 +4,9 @@
   <li>
     <sky-code>-t addin</sky-code> — References the <a href="https://github.com/blackbaud/skyux-sdk-template-addin">SKY UX Add-in template</a> to create <a href="https://developer.blackbaud.com/skyapi/docs/addins">SKY Add-ins</a> that allow developers to build features that integrate with Blackbaud solutions.
   </li>
+  <li>
+    <sky-code>-t barebones</sky-code> — References a pared down version of the <a href="https://github.com/blackbaud/skyux-sdk-template-barebones">SKY UX Add-in template</a> to create <a href="https://developer.blackbaud.com/skyapi/docs/addins">SKY Add-ins</a> that allow developers to build features that integrate with Blackbaud solutions. This template eliminates folders, modules, and references to packages that are not required to build add-ins.
```
blackbaud-johnly: The text for the first link seems off to me. I mean, the link text is the same as in the `-t addin` description, but it points to a different repo. That seems a little confusing to me. It's a minor thing, but what do you think about making "pared down version of the" part of the link text? I think that makes it clearer what we are linking to.

PR #745, opened 2019-11-14
Update skyAppConfig property chain to get value

zeiddev:
When I went to get a value from the skyAppConfig object the code example given was not correct. There was no params property directly beneath the skyAppConfig object. Instead I called the runtime property and then the params property (I assume that this was correct)


blackbaud-johnly:
Looks like checks won't pass on this PR, so I reproduced the change in https://github.com/blackbaud/skyux2-docs/pull/748 to get the checks to pass. Thanks for the contribution @zeiddev!

Issue #746, opened 2019-11-15
Document new repeater item property and repeater event


blackbaud-johnly:
See https://github.com/blackbaud/skyux-lists/pull/91.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/750.

PR #750, opened 2019-11-17
Document tag property and related event for repeaters

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/746. I assume this is good to go since we released https://github.com/blackbaud/skyux-lists/pull/91.

### DIFF
--- a/src/app/components/repeater/index.html
+++ b/src/app/components/repeater/index.html
@@ -41,6 +41,15 @@
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
+  <sky-demo-page-properties sectionHeading="Repeater events">
+    <sky-demo-page-property
+      propertyName="orderChange"
+      isOptional="true"
+    >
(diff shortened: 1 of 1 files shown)


src/app/components/repeater/index.html
```diff
@@ -89,7 +98,15 @@
     >
     Indicates whether to display an inline form within the repeater. Users can toggle between displaying and hiding the inline form. This property accepts <sky-code>boolean</sky-code> values.
     </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="tag"
+      defaultValue="-1"
```
Blackbaud-AlexKingman: There is no default value for `tag`. It's just `undefined`. Also, it accepts `any` values, not `number`.

blackbaud-johnly: Sorry. I got my wires crossed and was looking at `tabIndex` in the code. I'll update. Thanks!

PR #749, opened 2019-11-15
[HOLD] List view switcher docs

Blackbaud-ToddRoberts:

Issue #726, opened 2019-10-30
Document list view switcher component


Blackbaud-AlexKingman:
blackbaud/skyux-list-builder#32

PR #751, opened 2019-11-17
Doc tweaks to list view switcher docs

blackbaud-johnly:
Edits to the content at https://github.com/blackbaud/skyux2-docs/pull/749 and a first stab at the development side of https://github.com/blackbaud/skyux2-docs/issues/726 (minus the demo).


src/app/components/list/view-switcher/index.html
```diff
@@ -4,7 +4,7 @@
 
   <sky-demo-page-summary>
     <p>
-      The list view switcher lets the user switch between different views of the same data.
+      The list view switcher module displays toolbar options that enable users to switch between different views of the same data.
     </p>
   </sky-demo-page-summary>
```
blackbaud-johnly: @Blackbaud-AlexKingman, are you planning to create the demo for this component? Or should I look into trying to set that up? (I assume it'll be much quicker for a developer to create the demo, but I can work on it if necessary.)

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
       <h3 class="sky-text-success">Use when</h3>
         <sky-docs-visual-example>
           <sky-docs-visual-example-text>
-              <p>Use the list view switcher when the user needs to switch between different views of the same data.</p>
+              <p>Use list view switchers when users need to switch between different views of the same data.</p>
```
blackbaud-johnly: @Blackbaud-ToddRoberts, should we provide an example of when/why multiple views are necessary to provide some context for the use cases of this component?

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
           </sky-docs-visual-example-text>
           <sky-docs-visual-example-image>
             <div>
               <img src="~/assets/img/guidelines/list-view-switcher/use-when.png" />
```
blackbaud-johnly: @Blackbaud-ToddRoberts, the image here illustrates when two switcher buttons are in place, but it doesn’t illustrate the different views. Would it be useful to include a second image with the other button selected to show the difference in the views?

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
+              <div class="caption caption-success sky-text-success">Do use list view switchers when more than one view of data is available to users.</div>
             </div>
           </sky-docs-visual-example-image>
         </sky-docs-visual-example>
-      <h3 class="sky-text-success">Don't use when</h3>
+      <h3 class="sky-text-danger">Don't use when</h3>
         <sky-docs-visual-example>
           <sky-docs-visual-example-text>
-              <p>Don't use the list view switcher when there is only one view of the data.</p>
+              <p>Don't use list view switchers when only one view of data is available.</p>
```
blackbaud-johnly: @Blackbaud-ToddRoberts, is this Don’t use when item really necessary? It’s probably better to be explicit than to assume that users can/will read between the lines, but this item seems implicit based on the Use when item, so just making sure that we actually need it.

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
             </div>
           </sky-docs-visual-example-image>
         </sky-docs-visual-example>
     </sky-demo-page-content>
 
     <sky-demo-page-content sectionHeading="Anatomy">
       <sky-docs-anatomy imagePath="~/assets/img/guidelines/list-view-switcher/anatomy.png">
-        <sky-docs-anatomy-item anatomyNumber="1">Active view</sky-docs-anatomy-item>
-        <sky-docs-anatomy-item anatomyNumber="2">Inactive view</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="1">Active view button</sky-docs-anatomy-item>
```
blackbaud-johnly: @Blackbaud-ToddRoberts, JFYI that I added “button” to the item names for consistency with our naming for similar items in the datepicker, file attachment, and phone field Anatomy docs. Just calling this out in case you see any problem with this change.

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
             <div>
               <img src="~/assets/img/guidelines/list-view-switcher/dont-use-when.png" />
-              <div class="caption caption-danger sky-text-danger">Don't use the list view switcher when there is only one view of the data.</div>
+              <div class="caption caption-danger sky-text-danger">Don't use list view switchers when users cannot switch between different views.</div>
             </div>
           </sky-docs-visual-example-image>
         </sky-docs-visual-example>
     </sky-demo-page-content>
 
     <sky-demo-page-content sectionHeading="Anatomy">
```
blackbaud-johnly: @Blackbaud-ToddRoberts, it might just be me, but I was initially slightly confused by the icons and the active/inactive labels here. Since the icons here are the same as the icons in the other images, I thought for a minute that the first item here was a default button choice for displaying an "active view" and that the second one was a default button choice for displaying an "inactive view." I'm not sure if there's anything that needs to be addressed for this, but I just wanted to mention it in case you think other users might initially be similarly confused. The only tweak I can think of is to maybe use different icons in at least one of the Use when/Don't use when images. But I'm not sure if that's necessary; just throwing it out there.

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
+        <sky-docs-anatomy-item anatomyNumber="1">Active view button</sky-docs-anatomy-item>
+        <sky-docs-anatomy-item anatomyNumber="2">Inactive view button</sky-docs-anatomy-item>
       </sky-docs-anatomy>
     </sky-demo-page-content>
 
     <sky-demo-page-content sectionHeading="Options">
       <h3>Available views</h3>
         <div class="sky-text-block">
-          <p>You must specify which views are available for the user to switch between.</p>
+          <p>You must specify the views that are available to users.</p>
```
blackbaud-johnly: @Blackbaud-AlexKingman, how do users specify the views? Looking in the code, the only place where I see to specify anything is the custom button component. Is that where this is specified? And if so, how is this different than the custom views we mention two subsections down?

src/app/components/list/view-switcher/index.html
```diff
@@ -21,44 +21,44 @@
...
         </div>  
       <h3>Default view</h3>
         <div class="sky-text-block">
-          <p>You must specify the default view to show when the page loads.</p>
+          <p>You must specify the default view to display when the page loads.</p>
```
blackbaud-johnly: @Blackbaud-AlexKingman, where is the default view set? Looking at the code, I don’t see an option to designate one of the views as the default. Is this actually something that is set through the list view switcher module?

src/app/components/list/view-switcher/index.html
```diff
@@ -69,12 +69,12 @@ <h3>Custom views</h3>
     <sky-demo-page-content sectionHeading="Behavior & states">
       <h3>Switch view</h3>
```
blackbaud-johnly: @Blackbaud-ToddRoberts, should this subheader be “Filters”? This section seems to be about the behavior of filters, so that header seems more precises to me. Or am I misunderstanding what we are calling out here?

src/app/components/list/view-switcher/index.html
```diff
@@ -106,7 +106,30 @@ <h3>Components</h3>
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="view"
+      isOptional="true"
+    >
+      Specifies the view that the list view switcher button displays. This property accepts <sky-code>ListViewComponent</sky-code> properties.
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
+</sky-tab>
```
blackbaud-johnly: @Blackbaud-AlexKingman, are you planning to create the code samples for this component? Or should I look into trying to set that up? (Unlike the demos, I'm not sure it's really even an option for me to look into creating full code samples, but let me know if I should look into it..)

src/app/components/list/view-switcher/index.html
```diff
@@ -106,7 +106,30 @@ <h3>Components</h3>
     >
     </sky-demo-page-module-info>
 
-  </sky-tab>
+    <sky-demo-page-properties
```
blackbaud-johnly: @Blackbaud-AlexKingman, do we need to create any additional development reference docs for this component? These were the only public properties I saw when I scanned the code, but let me know if you were planning on any additional info here or if we need to add anything about the component that contains this one.

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: Probably extraneous but @Blackbaud-AdamFunderburk and I have been discussion whether we should always have "don't" examples - what do you think, Adam?

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: I think the issue right now is that there is only one built in mode to choose from

src/app/components/list/view-switcher/index.html
Blackbaud-AdamFunderburk: I'm a believer in having at least one example of how a component could plausibly be misapplied. It's not _bad_ to include this in my opinion - I don't think it's that implausible (based on my experience) for a consumer to mistakenly assume that it's a standard part of a list toolbar and should always be included, OR that it's ok to include it in a V1 release of a list feature because later versions will include additional views.

Just while I'm thinking about it, another "don't" example could be using the list view switcher to toggle between different "views" of a list, outside of the presentation options intended. A consumer could mistakenly use this (because it looks cool!) to accommodate different pre-built sort/filter options of the same list, when we would instead recommend using tabs, persistent inline filters, etc.

Blackbaud-AdamFunderburk:
At least in the version of the PR I'm seeing, the images in the States table are using borders, which they should not. Images in the states table do not need borders, and do not need to conform to 290/600 width requirements.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: Makes sense to me to err on the side of including at least one "don't" example. ... Do you guys want to add Adam's additional "don't" example? Or just leave it as the existing example?

blackbaud-johnly:
I see the borders on the States images too, but I assume we can address that in https://github.com/blackbaud/skyux2-docs/pull/749. I didn't add borders in my branch.

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: Discussed this with @Blackbaud-AdamFunderburk, our thought is that the demo should be sufficient for illustrating this.

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: added

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: i am inclined to leave it with the existing example

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: good call

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: Updated the header and text. Not sure if the header is exactly right but trying to convey that the data model is preserved when switching.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: Sounds good. Thanks!

src/app/components/list/view-switcher/index.html
blackbaud-johnly: LGTM. Thanks!

src/app/components/list/view-switcher/index.html
blackbaud-johnly: LGTM

src/app/components/list/view-switcher/index.html
blackbaud-johnly: Sounds good. Thanks.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: @Blackbaud-ToddRoberts, I assume that we don't need to worry about this?

Blackbaud-ToddRoberts: i think it's ok, we can update the image later if it is causing confusion.

blackbaud-johnly: Sounds good. Resolving this comment...

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: I can work on creating a demo.

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: So this is where things get tricky. Today, we do not document how to create custom views. I do know we have a couple consumers that dug into the code and are doing it but we've never documented it as I don't know how main stream we wanted to make it. The design documentation for the view switcher specifically called out supporting custom views though and so it does just that. I feel like documenting the view switcher at all isn't extremely helpful today if we don't also document the custom views (since there is only one supported view today) but I don't know what @Blackbaud-ToddRoberts thinks.

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: That property is on the list component not on the view switcher itself.

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: I'm not sure what you mean here by code sample vs demo? Your other question asked about the overall demo so I want to make sure I'm following you correctly. Are you looking for something different than that here?

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: The properties are correct but we probably should call out somewhere what this is for. You use this when using the view switcher with a custom view that does not have a built in configuration to show up automatically. (Currently only the grid shows up automatically. The plan was for repeater and cards to do the same but not sure about that now that we may be doing something different there). 

I think this goes back to my other comment regarding us figuring out what we want to document for custom views.

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: I think we need to document creating custom views. There is nothing inherently wrong with using them if we don't have a suitable built in view.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: I think I was getting a little ahead of myself. I noticed that we are missing the demo and code sample, so I wanted to make sure both are accounted for. The reason I was thinking of them as distinct is that in the new docs pattern, we have the responsive demo and code samples that expand on different scenarios and aren't just the code for the demo. But we aren't using the new docs pattern here, so my question here didn't make much sense. Sorry.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: Resolving this issue as we've decided to move forward with the docs minus the demo and circle back on the demo and code samples.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: I fleshed out the description of the `view` property to try to call out the point you made. Do you think this works?

src/app/components/list/view-switcher/index.html
blackbaud-johnly: @Blackbaud-ToddRoberts and @Blackbaud-TrevorBurch, where would we document custom views? I assume on the list docs as opposed to here. Is that right? If so, I think we need to create a separate issue to document custom views. Please let me know what you think.

src/app/components/list/view-switcher/index.html
blackbaud-johnly: I added a sentence here to call this out. Please let me know if you see any problems with it.

blackbaud-johnly:
@Blackbaud-AdamFunderburk @Blackbaud-ToddRoberts, and @Blackbaud-TrevorBurch, I went back through the docs and made a handful of updates to get these docs ready to go without the demo and code samples. We still have three unresolved comments, but I think two are addressed and just need confirmation. And I think we need to file a separate issue for the other one, but I'd like confirmation there too. If anyone can take a look at the comments and give the docs a once-over, I think we can get this merged into the parent branch this week and then hopefully get that PR out to update the docs site. Please just let me know if you see any problems or need anything else from me on this. Thanks!

src/app/components/list/view-switcher/index.html
Blackbaud-ToddRoberts: I'm not sure the best place. Will the method be changing with the new data manager?

Blackbaud-TrevorBurch: I would agree that the custom view work probably fits better on the main list topic. And as for the new data manager - this component is only for the old one. I'm not sure how we will want to resolve/replace current list stuff with that when it is available but I don't think we should attempt to solve that problem with this component.

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: LGTM

src/app/components/list/view-switcher/index.html
Blackbaud-TrevorBurch: LGTM

src/app/components/list/view-switcher/index.html
blackbaud-johnly: Great. I'm resolving this issue here and creating a separate docs issue at https://github.com/blackbaud/skyux2-docs/issues/825.

Issue #727, opened 2019-10-30
Clean up docs for icons


blackbaud-johnly:
Based on user feedback, update icon component docs to call out that the "fa fa-" classes should not be specified. Also suggested; in the Styles/Icons section section, show how to color the icon using one of the appropriate styles.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/752.

PR #752, opened 2019-11-18
Update icon docs

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/727. ... This addresses the point that an external developer raised about the `fa fa-` classes. But it does not include anything about how to color the icon using one of the appropriate styles as @Blackbaud-BenLambert suggested. Do we want to include anything about that? And if so, what guidance do we want to provide?

### DIFF
--- a/src/app/components/icon/index.html
+++ b/src/app/components/icon/index.html
@@ -12,7 +12,7 @@
     <sky-demo-page-property
       propertyName="icon"
     >
-      Specifies the name of the Font Awesome icon to display. This property accepts <sky-code>string</sky-code> values.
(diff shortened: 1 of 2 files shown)


Blackbaud-BenLambert:
IMO, I think it would be good to show an example of coloring the icon.  I've seen some inconsistent attempts at doing this, so I believe we need to include at least a very basic example of choice a <sky-icon> in something other than the default black color.

blackbaud-johnly:
@Blackbaud-BenLambert, I looked into this, and we don't currently have a first-class way to change the color of the icons. I talked this over with @Blackbaud-SteveBrush and @Blackbaud-BobbyEarl, and I created https://github.com/blackbaud/skyux-indicators/issues/45 for that dev work rather than documenting any workarounds for this.

Blackbaud-BenLambert:
ok thanks - I may be mistaken about this practice though, maybe it's not as common as I thought...glad for you guys to make the call on it either way.

PR #768, opened 2019-12-03
Error docs redesign

Blackbaud-AdamFunderburk:
I did NOT duplicate the existing demo at the top of the page, like we usually do for our Phase 1 redesigns. This component's demo in particular is enormous and pushes the tabbed content far enough down to be harmful. I'm open to suggestions, but my impression is that we should find another solution for the masthead demo for Error - either rushing the priority of its Phase 2 redesign or making an inexpensive replacement for the meantime.


src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+                  <img src="~/assets/img/guidelines/error/error-usage-7.png" />
+                    <div class="caption caption-success sky-text-success">Do use a toast to inform the user about a process operating in the background.</div>
+                </div>
+            </sky-docs-visual-example-image>
+
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+        <sky-docs-visual-example-text>
+            <p>Don’t use an error when the status of specific content needs user attention, but the system has not necessarily broken. To notify the user of a broken status on a workflow or record, use an <a routerLink="/components/alert">alert</a> instead.</p>
```
blackbaud-johnly: I'd change "an error" to "error messages." Also, it's probably just me, but I'd change "needs" to "requires." ... Also, is "the system has not necessarily broken" correct? I mean, it's not the "system" that breaks is it? And "not necessarily" is kinda ambiguous. I wonder if we can change the second half of this sentence to "... user attention but no error has occurred." (I'd like to tweak that to use active voice, but that's the best I've been able to do so far.

In the second sentence, I'd change "the user" to "users." And I'd change "an alert" to "alerts."

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+        </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div class="sky-margin-stacked-content-section">
+              <img src="~/assets/img/guidelines/error/error-usage-3.png" />
+                <div class="caption caption-danger sky-text-danger">Don't use the error component to represent issuse with content that user would expect to address.</div>
```
blackbaud-johnly: I'd change "the error component" to "error messages." ... Also, we've misspelled "issues," but I'm not sure that it's the right word here. What do you think of "Don't use error messages to highlight content that users must address when no error occurred."?

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+            </div>
+              <div>
+              <img src="~/assets/img/guidelines/error/error-usage-4.png" />
+                <div class="caption caption-success sky-text-success">Do use alert to call attention to issues that are intended to prompt user interaction.</div>
```
blackbaud-johnly: I'd change "alert" to "alerts." Also, should "alerts" be a link to that component topic? ... It's nitpicky, but the "issues" aren't really "intended to prompt user interaction." I'd change "are intended to prompt" to "require."

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+            </div>
+          </sky-docs-visual-example-image>
+                
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+        <sky-docs-visual-example-text>
+          <p>Don’t use the security error type when a user doesn’t have permission to view a portion of the page such as a tile on a record page. Hide the tile from view instead.</p>
```
blackbaud-johnly: I'd change "the security error type" to "security error messages" and change "a user doesn't have" to "users lack." And I'd change "a portion" to "portions." And based on the caption, it seems like we should mention modals in this sentence. I'd change the end of this sentence to "... portions of a page or modal such as a tile." And in the last sentence, I'd change "tile" to "content" to make a generic rule instead of a specific example.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+        </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div>
+              <img src="~/assets/img/guidelines/error/error-usage-5.png" />
+                <div class="caption caption-danger sky-text-danger">Don't use a security error in a portion of a page or modal. Hide the restricted content instead.</div>
```
blackbaud-johnly: I'd change "the security error type" to "security error messages" and change "a user doesn't have" to "users lack." And I'd change "a portion" to "portions."

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+    
+    </sky-demo-page-content>
+        
+    <sky-demo-page-content sectionHeading="Options">
+        
+        <h3>Error type</h3>
+        
+        <sky-docs-visual-example>
+            <sky-docs-visual-example-text>
+              <p>Use the broken error type in situations where the underlying content doesn’t load or when there is an issue reaching the server, such as 500 errors.</p>
```
blackbaud-johnly: I'd tweak the end of the sentence to: "... or when an issue such as a 500 error reaches the server."

Also, should `error` be a code snippet since it's referring to the type that is specified with the `errorType` property? I think we're probably fine not using a code snippet since this is a generic reference, but I figured it was worth asking. Same question for the other types.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+            <sky-docs-visual-example-image>
+                <div>
+                  <img src="~/assets/img/guidelines/error/error-option-broken.png" />
+                </div>
+            </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+            <sky-docs-visual-example-text>
+              <p>Use the construction error type when a page is under maintenance or is experiencing some sort of known outage.</p>
```
blackbaud-johnly: I'd change "is experiencing some sort of known" to "experiences a known."

Also, if we don't use code snippets for the types, I'd suggest changing "construction" to "under construction." If this is a reference to the property types, then a code snippet of `construction` makes sense. But for generic usage "under construction" makes more sense.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+            <sky-docs-visual-example-image>
+                <div>
+                  <img src="~/assets/img/guidelines/error/error-option-construction.png" />
+                </div>
+            </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+            <sky-docs-visual-example-text>
+              <p>Use the not found error type when there is an error loading a page due to a 404 error.</p>
```
blackbaud-johnly: I'd change the when clause to "when a page fails to load because of a 404 error."

Also, if we use code snippets for the error types, then `notfound` should be one word. But if we stick with generic references, it's fine as is.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+              <p>Use the security error type when a user doesn’t have access to view content.</p>
+            </sky-docs-visual-example-text>
+            <sky-docs-visual-example-image>
+                <div>
+                  <img src="~/assets/img/guidelines/error/error-option-security.png" />
+                </div>
+            </sky-docs-visual-example-image>
+        </sky-docs-visual-example>
+        
+        <h3>Image and action</h3>
```
blackbaud-johnly: Should image and action really be combined into a single section? I mean, aren't there instances where you'd use one and not the other? And if so, shouldn't they have separate guidelines? For example, in our guidelines at https://developer.blackbaud.com/skyux/design/guidelines/user-assistance#empty-state-help, the Invitations example includes an action but not an image. Since both are optional, it seems like they can be used separately, but the way this section is written makes it sound like they are a package deal. Also, do we need to include something about the description in the Options section? Seems like we should.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+        
+        <sky-docs-visual-example>
+            <sky-docs-visual-example-text>
+                <p>Include the optional image and action when the error component is the only content on the page. Tiles have limited space and may not be important to users' immediate tasks, so exclude the images and actions and only display error messages.</p>
```
blackbaud-johnly: I'd start with "Only include..." and delete "optional." ... And at the end of the second sentence, I think "error messages" should be "the title and description" to correspond with the anatomy items. I think the "error message" is the entire thing. Or am I thinking about this incorrectly?

Also, the Invitations image at https://developer.blackbaud.com/skyux/design/guidelines/user-assistance#empty-state-help is at odds with the guidance not to include actions in tiles, and I think I've seen several existing tiles where the empty state help includes calls to action. Do we really mean to go against the tide on this?

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+            </sky-docs-visual-example-text>
+            <sky-docs-visual-example-image>
+                <div>
+                  <img src="~/assets/img/guidelines/error/error-usage-6.png" />
+                    <div class="caption caption-success sky-text-success">Keep the error message more subtle when it affects only a portion of the content on a page.</div>
```
blackbaud-johnly: Is subtlety the driving focus here or space? I think I'd tweak this caption to something like "Exclude images and actions when error messages only affect portions of the content on a page."

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+        </sky-docs-visual-example>
+        
+    </sky-demo-page-content>
+        
+    <sky-demo-page-content sectionHeading="Content">
+        
+        <h3>Style and tone</h3>
+        
+            <div class="sky-text-block">
+                <p>Make error and warning messages brief and as accurate as possible to help the user address the problem, and use a conversational approach. When possible, offer a solution along with an explanation of why the error or warning is occurring. Use the error component to provide a consistent user experience with error messages.</p>
```
blackbaud-johnly: Why "and warning"? Up to this point, the terminology has been "error" only. Seems weird to alter the terminology at this point. Also, what are we trying to say with "accurate as possible"? The implication is that the error messages won't always be accurate, which I don't think is what we mean to say here. Also, the first and second sentences both state similar points about solutions. I'm not exactly sure what we mean to say here, but I imagine this should be something like "Make error messages brief, and use a conversational tone. Explain why the error occurred, and when possible, offer a solution to help users address the problem." ... And do we need the third sentence? I'm not sure what it's adding here.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+                <p>Errors can frustrate users and sometimes indicate serious problems, so be thoughtful and helpful with the wording. Limit the use of "please" and "sorry," be careful about how you use "can't" and "don't" (avoid "you can't" and "you don't"), and avoid humor.</p>
```
blackbaud-johnly: I'd remove the conditionals here: "Errors frustrate users and indicate serious problems, so error messages must be thoughtful and helpful."

And in the second sentence, I'd tweak the order so the "avoid humor" doesn't get lost at the end because that reads awkwardly. And the middle item is a little vague, so I'd edit it to be more direct: "Avoid humor, limit the use of 'please' and 'sorry,' and avoid negative constructions such as 'you can't' and 'you don't.' "

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+    <sky-demo-page-content sectionHeading="Related information">
+        
+        <h3>Components</h3>
+          <ul>
+            <li><a routerLink="/components/alert">Alert</a></li>
+            <li><a routerLink="/components/Toast">Toast</a></li>
+          </ul>
+        <h3>Guidelines</h3>
+          <ul>
+            <li><a routerLink="/design/guidelines/error-handling">Error handling</a></li>
```
blackbaud-johnly: Seems like we should also have a link to the User assistance guidelines at https://developer.blackbaud.com/skyux/design/guidelines/user-assistance since that comes into play here.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
+          </ul>
+        
+    </sky-demo-page-content>    
+    
+    </sky-tab>
+    
+    <sky-tab tabHeading="Development">
+
```
blackbaud-johnly: We are missing the installation section at the top of the Development tab.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
-<sky-demo-page
+<sky-tabbed-demo-page
   pageTitle="Error"
-  skyuxModule="SkyErrorModule"
-  packageName="@skyux/errors"
 >
 
-  <sky-demo-page-summary>
+  <sky-demo-page-summary class="sky-margin-stacked-content-section">
     The error component displays a SKY UX-themed error message. It includes options to specify an image, title, description, and action in the error message.
```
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: I'm fine with this change, but on reflection I'm more in favor of just removing this second sentence entirely. What do you think?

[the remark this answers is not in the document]
blackbaud-johnly: That makes sense to me.

src/app/components/error/index.html
```diff
@@ -1,13 +1,196 @@
...
   </sky-demo-page-summary>
```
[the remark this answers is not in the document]
Blackbaud-AdamFunderburk: An interesting approach. That wording might be a little redundant - can you recommend how to condense it?

src/app/components/error/index.html
Blackbaud-AdamFunderburk: How about
"Don’t use error messages when the status of specific content requires user attention, but there is no system error. To notify users of a broken status on a workflow or record, use an <a routerLink="/components/alert">alerts</a> instead."

src/app/components/error/index.html
Blackbaud-AdamFunderburk: I think we should avoid code snippets in the "Design" tab, to avoid discussing specific implementation when we should be discussing behavior.
BUT, would it be worthwhile to put the type name itself in quotations?

> Use the "broken" error type in situations where...

src/app/components/error/index.html
Blackbaud-AdamFunderburk: I would consider this difference an intentional distinction between "error handling" and "empty state". The recommended guidelines for prompting user action (with a button) in cases where records are incomplete or could be enhanced feels like it matches our principles well. By contrast, when a portion of page content fails to load because of a system error, prompting user action with the error component's primary (!!) action button doesn't match our principles well.

Maybe we should spell out that distinction here with an example, and link to the empty state guidelines as a way of clarifying that point.

src/app/components/error/index.html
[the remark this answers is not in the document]
blackbaud-johnly: What do you think of deleting the first sentence and breaking up the second sentence into 2 shorter sentences: 
"We are creating an error component demo to fit in this section. To view the existing demo in the meantime, select the Development tab."

src/app/components/error/index.html
blackbaud-johnly: Looks good to me.

src/app/components/error/index.html
blackbaud-johnly: We could discuss that pattern of quote marks, but my initial instinct is to not use them in this scenario. I think it would add ambiguity, but it's worth talking to Todd and others if you think it would make things more clear. Personally, I'd lean toward generic references like you have here without the quote marks. When I noticed that the types corresponded to the properties on the Development tab, I just wanted to make sure we were intentionally not using code snippets.

src/app/components/error/index.html
blackbaud-johnly: Seems like I was getting my wires crossed here and inferring a pattern conflict when it's really 2 distinct patterns. Sorry about that. If you think this is something that might confuse others too, than I agree we could spell out the distinction, but feel free to let me know if you think it's unlikely to be necessary. If we do want to address this, maybe we could add a third sentence to the end along the lines of "This pattern is different than empty state help, which can include actions." And we could make "empty state help" the link to those guidelines. What do you think?

Also, just to follow up on the suggestion to start with "Only include..." and delete "optional." .... As written, it sounds like the guideline is to ALWAYS include the image and action in error messages on pages. If that's the point we are making, I'd split this into 2 paragraphs after the first sentence to make it clearer that the first sentence is a distinct point about pages and that the rest is a distinct point about tiles and such. But if the point we are trying to make is not that images and actions are always required on pages but rather that they should NOT be used in tiles and such, I'd start with "Only include..." and delete "optional."

blackbaud-johnly: The updates here look good to me!

src/app/components/error/index.html
```diff
@@ -1,13 +1,232 @@
...
 
+        </sky-docs-visual-example>
+        
+        <sky-docs-visual-example>
+        <sky-docs-visual-example-text>
+            <p>Don’t use error messages when the status of specific content requires user attention, but there is no system error. To notify users of a broken status on a workflow or record, use <a routerLink="/components/alert">alerts</a> instead.</p>
+        </sky-docs-visual-example-text>
+          <sky-docs-visual-example-image>
+            <div class="sky-margin-stacked-content-section">
+              <img src="~/assets/img/guidelines/error/error-usage-3.png" />
```
Blackbaud-ToddRoberts: it took me a minute to figure out what was going on in this example. it may be helpful to show the full page context.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign](https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign)

src/app/components/error/index.html
```diff
@@ -1,13 +1,232 @@
...
+        </div>
+        
+    </sky-demo-page-content>
+        
+    <sky-demo-page-content sectionHeading="Related information">
+        
+        <h3>Components</h3>
+          <ul>
+            <li><a routerLink="/components/alert">Alert</a></li>
+            <li><a routerLink="/components/Toast">Toast</a></li>
```
blackbaud-johnly: The link here is busted. I'm pretty sure we just need to lowercase the "T" in "toast" at the end of the path.

src/app/components/error/index.html
```diff
@@ -1,13 +1,232 @@
...
+        
+    </sky-demo-page-content>
+        
+    <sky-demo-page-content sectionHeading="Content">
+        
+        <h3>Style and tone</h3>
+        
+            <div class="sky-text-block">
+                <p>Make error messages brief and use a conversational tone. Explain why the error occurred, and when possible, offer a solution to help users address the problem.</p>
+                <p>Errors frustrate users and indicate serious problems, so error messages must be thoughtful and helpful. Avoid humor, limit the use of "please" and "sorry", and avoid negative constructions such as "you can't" and "you don't."</p>
```
blackbaud-johnly: In the second sentence, the comma after "sorry" should be inside the quote marks, not outside of them.

src/app/components/error/index.html
blackbaud-johnly: I fixed this.

src/app/components/error/index.html
blackbaud-johnly: I fixed this.

src/app/components/error/index.html
blackbaud-johnly: Is this still an issue? It looks to me like the images now show a full-page context, but just making sure.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign](https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign)

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign](https://developer.blackbaud.com/skyux?_sp.spa=pr-error-docs-redesign)

Issue #271, opened 2018-09-18
Document the ability for libraries to include an isolated "testing" module


blackbaud-johnly:
See https://github.com/blackbaud/skyux-builder/pull/471.

Blackbaud-DeniseCisneros:
Fixed and merged in https://github.com/blackbaud/skyux2-docs/pull/775

PR #775, opened 2019-12-09
Added test module section

Blackbaud-DeniseCisneros:
References issue https://github.com/blackbaud/skyux2-docs/issues/271


src/app/learn/reference/component-libraries/index.html
```diff
@@ -177,4 +177,24 @@
     }
   </sky-code-block>
 
+  <stache-page-anchor>
+    Test modules
+  </stache-page-anchor>
+
+  <p>
+    You can generate a separate module for testing by adding a test directory. The test module uses a separate test file to prevent cluttering the production module during webpack builds.
```
blackbaud-johnly: In the first sentence, should the subject be "Libraries" instead of "You"? I _think_ so in the sense that "you" set up directory but that the library itself seems to generate the module. Also, it seems like "an isolated" might be more precise and meaningful than "a separate." Also, Steve noted in the PR for this feature that consuming SPAs use the testing module for unit tests. It seems like maybe we should include this point. (I'd also capitalize "Webpack.") What do you think of rearranging like this:

"Libraries can generate isolated test modules for consuming SPAs to use in unit tests. The testing module avoids cluttering the production module during Webpack builds."

src/app/learn/reference/component-libraries/index.html
```diff
@@ -177,4 +177,24 @@
...
+  </p>
+
+  <sky-code-block>
```
blackbaud-johnly: This code block seems unnecessary. It seems like this should just be a code snippet combined with the next paragraph. What do you think of:
"To create a testing module, add a `testing` directory to `src/app/public` and then add an `index.ts` file. The library exports the contents of the `index.ts` file to NPM as a testing module."

src/app/learn/reference/component-libraries/index.html
```diff
@@ -177,4 +177,24 @@
...
+    src/
+      app/
+        public/
+          testing/
+            index.ts
+  </sky-code-block>
+
+  <p>
+    The contents of the <sky-code>index.ts</sky-code> file export as a test module to NPM.
```
blackbaud-johnly: See note above about maybe combining this with the the code block after converting the code block to code snippets.

Issue #694, opened 2019-10-10
Document that highlight directive can't be placed on a template element.


Blackbaud-TrevorBurch:
For blackbaud/skyux-indicators#36

The highlight direct can not be placed on `ng-template`, `ng-content` or `ng-container` tags. It must be placed on elements which truly have content.

Blackbaud-DeniseCisneros:
Fixed and closed in https://github.com/blackbaud/skyux2-docs/pull/777

PR #777, opened 2019-12-11
Added note about text highlight directive

Blackbaud-DeniseCisneros:
Added note about where the text highlight directive can be placed. References https://github.com/blackbaud/skyux2-docs/issues/694

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -544,9 +544,9 @@
       }
     },
     "@pact-foundation/pact-node": {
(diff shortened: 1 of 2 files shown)


Blackbaud-DeniseCisneros:
@Blackbaud-TrevorBurch Based on the comment from Christi in https://github.com/blackbaud/skyux-indicators/issues/36, will the demo need updated at all or should the added note be fine?

PR #827, opened 2020-02-03
Document fluid-grid gutterSize and disableMargin properties

Blackbaud-AlexKingman:
Addresses #826

### DIFF
--- a/src/app/components/fluid-grid/index.html
+++ b/src/app/components/fluid-grid/index.html
@@ -9,6 +9,34 @@
     The fluid grid component provides a responsive 12-column layout to organize content for all device sizes. The fluid grid consists of the <sky-code>sky-fluid-grid</sky-code>, <sky-code>sky-row</sky-code>, and <sky-code>sky-column</sky-code> tags. You place <sky-code>sky-column</sky-code> elements in <sky-code>sky-row</sky-code> elements, and then you can specify how much of the layout to use for each <sky-code>sky-column</sky-code> element on extra-small, small, medium, and large screens. The <sky-code>sky-fluid-grid</sky-code> parent element ensures consistent styling for all fluid grid instances.
   </sky-demo-page-summary>
 
+  <sky-demo-page-properties sectionHeading="Fluid grid properties">
+    <sky-demo-page-property
+      propertyName="disableMargin"
+      defaultValue="false"
+      isOptional="true"
+    >
+      Disables the outer left and right margin of the fluid grid container. This property accepts <sky-code>boolean</sky-code> values.
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="gutterSize"
+      defaultValue="SkyFluidGridGutterSize.Large"
+      isOptional="true"
+    >
+      Specifies a <sky-code>SkyFluidGridGutterSize</sky-code> enum that defines the size of the padding between each column. The valid options are:
+      <ul>
+        <li>
+          <sky-code>Small</sky-code> — Defines a 10px gutter.
+        </li>
+        <li>
+          <sky-code>Medium</sky-code> — Defines a 20px gutter.
+        </li>
+        <li>
+          <sky-code>Large</sky-code> — Defines a 30px gutter.
+        </li>
+      </ul>
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
   <sky-demo-page-properties sectionHeading="Row properties">
     <sky-demo-page-property
       propertyName="reverseColumnOrder"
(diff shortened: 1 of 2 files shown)


Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin](https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin)

src/app/components/fluid-grid/index.html
```diff
@@ -9,6 +9,34 @@
     The fluid grid component provides a responsive 12-column layout to organize content for all device sizes. The fluid grid consists of the <sky-code>sky-fluid-grid</sky-code>, <sky-code>sky-row</sky-code>, and <sky-code>sky-column</sky-code> tags. You place <sky-code>sky-column</sky-code> elements in <sky-code>sky-row</sky-code> elements, and then you can specify how much of the layout to use for each <sky-code>sky-column</sky-code> element on extra-small, small, medium, and large screens. The <sky-code>sky-fluid-grid</sky-code> parent element ensures consistent styling for all fluid grid instances.
   </sky-demo-page-summary>
 
+  <sky-demo-page-properties sectionHeading="Fluid grid properties">
+    <sky-demo-page-property
+      propertyName="disableMargin"
+      defaultValue="false"
+      isOptional="true"
+    >
+      Disables the outer left and right margin of the fluid grid container. This property accepts <sky-code>boolean</sky-code> values.
```
blackbaud-johnly: Do we need "outer" here? Is there an "inner" margin that we are distinguishing from? ... Also, should "margins" be plural? This reads to me like a left margin and a right margin, which suggests it should be plural. But I'm not sure.

src/app/components/fluid-grid/index.html
```diff
@@ -9,6 +9,34 @@
...
+    </sky-demo-page-property>
+    <sky-demo-page-property
+      propertyName="gutterSize"
+      defaultValue="SkyFluidGridGutterSize.Large"
+      isOptional="true"
+    >
+      Specifies a <sky-code>SkyFluidGridGutterSize</sky-code> enum that defines the size of the padding between each column. The valid options are:
```
blackbaud-johnly: It's superminor, but I'd change "that defines" to "to define" and change "each column" to "columns."

Also, do we really need the bullet list? If we really want the values to stand out, the list works. But we don't usually list values this way. Generally, we just list them in a sentence in the description. So in this case, I'm thinking something like: "The valid options are `Small` for a 10px gutter, `Medium` for a 20px gutter, and `Large` for a 30px gutter." ... We can keep the list if you prefer, but I don't think we really need the values to stand out that much.

src/app/components/fluid-grid/index.html
Blackbaud-AlexKingman: Yeah plural is fine. I think I added "outer" just to distinguish the difference between this and the gutters. I'm not sure if there is some better standardized language around these parts of the grid that would be a better fit?

src/app/components/fluid-grid/index.html
Blackbaud-AlexKingman: I just grabbed the list from another component. I like the sentence much better. I'll make that change.

Blackbaud-AlexKingman: done.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin](https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin)

src/app/components/fluid-grid/index.html
blackbaud-johnly: I updated this to use `sky-code` elements instead of tick marks. Otherwise, LGTM.

src/app/components/fluid-grid/index.html
blackbaud-johnly: I updated this to make "margins" plural. ... As for "outer," I'd just leave it. I don't know of any better standardized language. I just wasn't sure if we needed "outer" given that it seemed like the end of the sentence makes it clear that we are referring to the margin of the container, not the individual columns. And since `gutterSize` description sounds like that padding is only for in between columns (as opposed to the left of the leftmost column and the right of the rightmost column), I thought "outer" might be extraneous. But better to err on the side of leaving it in for clarity than deleting it to tighten up.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin](https://developer.blackbaud.com/skyux?_sp.spa=pr-fluid-grid-gutter-margin)

PR #832, opened 2020-02-10
Add date pipe docs

blackbaud-johnly:


Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-add-date-pipe-docs](https://developer.blackbaud.com/skyux?_sp.spa=pr-add-date-pipe-docs)

src/app/components/date-pipe/index.html
```diff
@@ -0,0 +1,3 @@
+<sky-docs-url-redirect
```
Blackbaud-BobbyEarl: Do we need a redirect here if it's never existed?

src/app/components/demo-components.service.ts
```diff
@@ -195,6 +195,12 @@ export class SkyDemoComponentsService {
         getCodeFiles: () => this.getDemoFiles('Copy to clipboard'),
         disableStackblitz: true
       },
+      {
+        name: 'Date pipe',
+        icon: 'calendar',
+        summary: `The date pipe formats date values according to locale rules.`,
+        url: '/components/date-pipe'
```
Blackbaud-BobbyEarl: Same question as above, I believe it's safe to ignore adding here as this service is no longer used.  I can confirm that though, and if it's true, probably remove this service all together.

blackbaud-johnly: Good point. Thanks for the offline explanation; I was confused and thought we needed the `date-pipe` folder in `skyux2-docs` to make this work. I'm closing this PR and making the updates we discussed in the `sky-skysp-svc` repo.

Issue #877, opened 2020-03-24
Update date range picker error message in demo


blackbaud-johnly:
Update message based on consumer feedback to remove switching from referring to From/To dates in field labels to start/end dates in the error message.

blackbaud-johnly:
Addressed in https://github.com/blackbaud/skyux2-docs/pull/878.

PR #878, opened 2020-03-24
Update date range picker demo error messages

blackbaud-johnly:
Addresses https://github.com/blackbaud/skyux2-docs/issues/877.

### DIFF
--- a/src/app/demos/date-range-picker/date-range-picker-demo.component.html
+++ b/src/app/demos/date-range-picker/date-range-picker-demo.component.html
@@ -19,7 +19,7 @@
         <div *ngIf="reactiveRange.errors.skyDateRange?.errors.endDateBeforeStartDate"
           class="sky-error-label"
         >
-          Select a To date that occurs after the From date.
+          Select a to date that occurs after the from date.
         </div>
         <div *ngIf="reactiveRange.errors.skyDate"
           class="sky-error-label"
--- a/src/app/demos/date-range-picker/date-range-picker-demo.component.html
+++ b/src/app/demos/date-range-picker/date-range-picker-demo.component.html
@@ -19,12 +19,12 @@
         <div *ngIf="reactiveRange.errors.skyDateRange?.errors.endDateBeforeStartDate"
           class="sky-error-label"
         >
-          End date must be before start date.
+          Select a To date that occurs after the From date.
         </div>
         <div *ngIf="reactiveRange.errors.skyDate"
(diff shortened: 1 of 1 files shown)


Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-edit-date-range-picker-error-032420](https://developer.blackbaud.com/skyux?_sp.spa=pr-edit-date-range-picker-error-032420)

src/app/demos/date-range-picker/date-range-picker-demo.component.html
```diff
@@ -19,12 +19,12 @@
         <div *ngIf="reactiveRange.errors.skyDateRange?.errors.endDateBeforeStartDate"
           class="sky-error-label"
         >
-          End date must be before start date.
+          Select a To date that occurs after the From date.
```
Blackbaud-TrevorBurch: Should we have quotes around "To date" and "From date"?

blackbaud-johnly: I went back and forth on how to handle this, and opted for this, which I don't love but like better than a lot of the alternatives. ... I think what we actually need to do here is lowercase `to` and `from` to make the them generic references to the dates, not specific references to the field labels. If we try formatting `To date` and `From date` with bold text or quote marks, then we'd also need to add a noun like `value` after those labels, so I think lowercasing `to` and `from` is the most practical solution. What do you guys think?

Blackbaud-DeniseCisneros: I agree that we should lowercase `to` and `from` so we're not referencing the specific labels.

Just thinking out loud and we don't have to solve this here, but I wonder if changing the component to use `Start` instead of `From` and `End` instead of `To` would make things clearer.

blackbaud-johnly: Yeah, I'd initially looked at updating the field labels, not the error message, but those are set by the component, not the demo, so I backed off on that. ... I'm going to go ahead and lowercase `to` and `from`. We can always revisit this later if necessary.

Blackbaud-BobbyEarl:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-edit-date-range-picker-error-032420](https://developer.blackbaud.com/skyux?_sp.spa=pr-edit-date-range-picker-error-032420)

PR #941, opened 2020-06-10
Ag grid updates

Blackbaud-ErikaMcVey:


src/app/components/data-entry-grid/index.html
```diff
@@ -113,6 +113,11 @@ <h2>ag-Grid styles</h2>
         To add the SKY UX styles for ag-Grid to your SPA, include <sky-code>"@skyux/ag-grid/css/sky-ag-grid.css"</sky-code> in the <sky-code>styles</sky-code> section of your <sky-code>skyuxconfig.json</sky-code>. For editable grids, include the class <sky-code>sky-ag-grid-editable</sky-code> on the <sky-code>ag-grid-angular</sky-code> element.
       </p>
 
+      <h2>ag-Grid wrapper component</h2>
+      <p>
+        The <sky-code>sky-ag-grid-wrapper</sky-code> component should wrap all uses of <sky-code>ag-grid-angular</sky-code>. The wrapper provides WCAG adherent keyboard navigation and sticky column headers for grids using auto height. It does not constrain standard ag-Grid functionality.
```
Blackbaud-DeniseCisneros, line 118: I'd move the explanation of what the wrapper does to the first sentence. Also, maybe compliant instead of adherent?

The <sky-code>sky-ag-grid-wrapper</sky-code> component provides WCAG compliant keyboard navigation and sticky column headers for grids using auto height. Use the <sky-code>sky-ag-grid-wrapper</sky-code> component to wrap all uses of <sky-code>ag-grid-angular</sky-code> without constraining standard ag-Grid functionality.

src/app/components/data-entry-grid/index.html
```diff
@@ -233,6 +238,13 @@ <h2>Passing cell editor parameters</h2>
         You can pass inputs to the SKY UX components used in cell editors via the <a href="https://www.ag-grid.com/javascript-grid-column-properties/">column definition's <sky-code>cellEditorParams</sky-code> property.</a> The component properties are nested under <sky-code>skyComponentProperties</sky-code>. All component properties should be supported, and you can see the defined types by importing <sky-code>SkyCellEditorDatepickerProperties</sky-code> or <sky-code>SkyCellEditorAutocompleteProperties</sky-code>. <sky-code>cellEditorParams</sky-code> can also be a function that returns a param object for dynamic editing params. See the demo for examples of using an object or a function.
       </p>
 
+      <h2>Using other SKY UX components in columns</h2>
+      <p>
+        There are some occasions where you need full control over the SKY UX component being rendered in the cell. In this case, you can create your own <a href="https://www.ag-grid.com/javascript-grid-cell-rendering-components/">cell renderer</a> and place a SKY UX component inside of it. Here are some cases where you will want to do this:
```
Blackbaud-DeniseCisneros, line 243: If you need full control over the SKY UX component rendered in a cell, create your own <a href="https://www.ag-grid.com/javascript-grid-cell-rendering-components/">cell renderer</a> and place the SKY UX component inside of it.

src/app/components/data-entry-grid/index.html
```diff
@@ -233,6 +238,13 @@ <h2>Passing cell editor parameters</h2>
...
+      </p>
+      <ul>
+        <li>Context menu - Creating your own context menu component gives you full control over the text, menu options, and click actions. See the demo for an example.</li>
```
Blackbaud-DeniseCisneros, line 246: I'd remove "Here are some cases..." and start a new line instead with the suggestion below. We can always go back to a bulleted list if more items are added, but we can ditch the list for now because there's only one example listed.

For example, use a cell renderer when creating your own context menu component for full control of the text, menu options, and click actions.

Blackbaud-ErikaMcVey: So the big goal for this section is to call out that you _need_ to make your own context menu. I think a lot of people expect it to be provided for them, but it is too complex of a component to make one that works for everyone. They don't see it in the docs, so they assume that ag-grid just "doesn't have context menu". The reason I made the list is because I think it draws the attention to that a bit more, instead of just being stuck in a paragraph that gets skimmed over. I also think there will be more components added to that list soon. 

Do you think there is a way to make it more clear that you can use a context menu, you just have to make it yourself? Or just putting it in that blurb will suffice?

Blackbaud-DeniseCisneros: It still feels a little off to use a bulleted list for one item, but I get wanting to quickly draw attention to it. Because that list will grow soon, that makes me more inclined to keep the list. I'll let @blackbaud-johnly weigh in too.

If we switch back to original format, is saying something like, "You must create your own context menu to fully control the text, menu options..." too strong of a prescription? In the original, I didn't quite take away that you _needed_ to but should.

blackbaud-ado:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-ag-grid-updates](https://developer.blackbaud.com/skyux?_sp.spa=pr-ag-grid-updates)

src/app/components/data-entry-grid/index.html
```diff
@@ -233,6 +238,13 @@ <h2>Passing cell editor parameters</h2>
         You can pass inputs to the SKY UX components used in cell editors via the <a href="https://www.ag-grid.com/javascript-grid-column-properties/">column definition's <sky-code>cellEditorParams</sky-code> property.</a> The component properties are nested under <sky-code>skyComponentProperties</sky-code>. All component properties should be supported, and you can see the defined types by importing <sky-code>SkyCellEditorDatepickerProperties</sky-code> or <sky-code>SkyCellEditorAutocompleteProperties</sky-code>. <sky-code>cellEditorParams</sky-code> can also be a function that returns a param object for dynamic editing params. See the demo for examples of using an object or a function.
       </p>
 
+      <h2>Using other SKY UX components in columns</h2>
```
blackbaud-johnly, line 241: Do we need `Using other` in this header? Or could it just be `SKY UX components in columns`?

Also, is `columns` right? The sentence that follows says `cells`, so should this be `cells` too?

src/app/components/data-entry-grid/index.html
```diff
@@ -233,6 +238,13 @@ <h2>Passing cell editor parameters</h2>
...
+      <p>
+        If you need full control over the SKY UX component rendered in a cell, you can create your own <a href="https://www.ag-grid.com/javascript-grid-cell-rendering-components/">cell renderer</a> and place the SKY UX component inside of it. Here are some cases where you will want to do this:
```
blackbaud-johnly, line 243: Can we change `If you need` to `For`? just to tighten up a little? Also, change `the` to `a` in front of `SKY UX`. I'd also delete `you can` to make this sound less conditional. And I'd delete the second instance of `SKY UX` to tighten up and avoid the repetition. And I'd delete 'of` after `inside` because it's just an extra word.

I'd also delete the "Here are some cases...` sentence as Denise suggested. See my note below for more info on that.

src/app/components/data-entry-grid/index.html
blackbaud-johnly: I'd delete the "Here are some cases...` sentence and also delete the bullet for now as Denise suggested. As she pointed out, we can always convert back to a bullet list when we have more items. I get the point that we intend to add more bullets, but I'd wait until we have those additional items before making this a bullet list.

As for the point about users _needing_ to make their own context menu, I agree with Denise that I don't get that point from the existing docs. I suggested an update to the previous paragraph to make this clear generically, so hopefully that works.

As for the content in the existing bullet, what do you think about rephrasing this a bit to emphasize the point about needing to make their own context menu component and state that it is an example so that we imply that the same is true for other components? I'm thinking of something like, `For example, to include a context menu with full control over text, menu options, and click actions, you create a context menu component and cell renderer and place the component in the cell renderer. See the demo for an example.` What do you think?

src/app/components/data-entry-grid/index.html
Blackbaud-ErikaMcVey: I think the "Using other" is important. This section comes after a section that talks about the SKY components we include editors and renderers for in the library. This section is to communicate that even though a SKY component might not have an editor provided by us, you can still use any component you want. Editors/renderers are set column-by-column, so every cell in a column will have the same editor/renderer.

blackbaud-ado:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-ag-grid-updates](https://developer.blackbaud.com/skyux?_sp.spa=pr-ag-grid-updates)

src/app/components/data-entry-grid/index.html
blackbaud-johnly: Keeping `Using other` makes sense. I was looking at the section in isolation and didn't make the connection.

PR #1003, opened 2020-08-25
Added autoClose docs & note

Blackbaud-DeniseCisneros:
References https://github.com/blackbaud/skyux-toast/issues/53

I wanted to bring attention to the note, but let me know if we should use something different than the alert or if I should use a different alert type.

### DIFF
--- a/src/app/components/toast/index.html
+++ b/src/app/components/toast/index.html
@@ -34,6 +34,20 @@
     </sky-demo-page-property>
   </sky-demo-page-properties>
 
+  <sky-demo-page-properties
+    sectionHeading="SkyToastComponent"
+  >
+    <p>
+      Selector
+    </p>
+    <sky-demo-page-property
+      propertyName="autoClose"
+      isOptional="true"
+    >
+      Indicates whether to automatically close the toast. <sky-alert alertType="info"> We do not recommend automatically closing toasts unless users can access the message after the toast is closed.</sky-alert>
+    </sky-demo-page-property>
+  </sky-demo-page-properties>
+
   <sky-demo-page-properties sectionHeading="SkyToastConfig type">
     <p>
       Specifies the configuration options to set up a toast.
--- a/src/app/components/toast/index.html
+++ b/src/app/components/toast/index.html
@@ -97,6 +97,9 @@
     <p>Toasts provide lightweight, time-sensitive information that is relevant to users when it appears. They guide users on actions to take and should include links for those actions whenever possible.</p>
     <p>The text must be clear and concise, and punctuation is subjective depending on the scenario.</p>
 
+    <div class="sky-subsection-heading">Behavior</div>
+		  <p>If toasts are the only place users can see the message, display the toasts until the user manually closes them. Only close toasts automatically if users can access the messages after the toasts close.</p>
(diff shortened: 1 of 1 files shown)


blackbaud-ado:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose](https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose)

src/app/components/toast/index.html
```diff
@@ -34,6 +34,20 @@
...
+    sectionHeading="SkyToastComponent"
+  >
+    <p>
+      Selector
+    </p>
+    <sky-demo-page-property
+      propertyName="autoClose"
+      isOptional="true"
+    >
+      Indicates whether to automatically close the toast. <sky-alert alertType="info"> We do not recommend automatically closing toasts unless users can access the message after the toast is closed.</sky-alert>
```
blackbaud-johnly, line 47: Just my 2 cents, but the alert feels too heavy to me. I'd just make it plain text. At one point Todd recommended a pattern of just using [status indicators](https://developer.blackbaud.com/skyux-indicators/docs/status-indicator) at the start of sentences to avoid alerts, but that never really caught on. I guess we could try that if you think we really need a visual cue to highlight this, but I'd argue we don't do that in cases where we include similarly important info. For example, the `iconType` property on action button calls out the supported version of SKY UX, which strikes me as similarly important since icons from later versions won't display, but that note is just part of the property description without any highlight:
image

Also, for whatever reason, looking at the note reminded me of a rule of thumb Todd advocated a while back where users respond better to guidance on what to do rather than what NOT to do. I wonder if we could reword this to follow that guidance:
```We recommend automatically closing toasts only when users can access the messages after the toasts close.```
Or even more emphatic:
```Only close toasts automatically if users can access the messages after the toasts close.```
(I feel like the pot calling the kettle out on that advice because I know that I forget about all the time and state warnings in negative terms. But I still think it's a good best practice.)

blackbaud-johnly: Oh, and we should also update the docs at https://developer.blackbaud.com/skyux-toast/docs/toast since I just merged the JSDocs for toast and we don't want to lose this note in the shuffle whenever the `docs-tools` version of the docs go live.

Blackbaud-DeniseCisneros: Ooo, good call on the guidance for the alert. I was sifting through the docs and trying to find examples of something similar in the actual property description, so I opted for taking my best guess and calling out the need for guidance in the description, ha! Updated with your feedback.

And yup, adding it to the JSDocs is a good idea so we don't lose it from this version to that one.

blackbaud-ado:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose](https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose)

blackbaud-ado:
Preview build at [https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose](https://developer.blackbaud.com/skyux?_sp.spa=pr-update-toast-autoclose)

PR #1005, opened 2020-08-25
Rc 4.0.0

Blackbaud-DeniseCisneros:
This content was on hold until the release of SKY UX 4; the content should be live now.

### DIFF
--- a/src/app/components/navbar/index.html
+++ b/src/app/components/navbar/index.html
@@ -5,7 +5,12 @@
 >
 
   <sky-demo-page-summary>
-    The navbar component displays a list of top-level navigation items. Navigation items can include behavior such as displaying sub-navigation items in dropdown menus with the <a routerLink="../dropdown">dropdown component</a>.
+    <p>
+      The navbar component displays a list of top-level navigation items. Navigation items can include behavior such as displaying sub-navigation items in dropdown menus with the <a routerLink="../dropdown">dropdown component</a>.
+    </p>
+    <p>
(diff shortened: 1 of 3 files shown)


src/app/learn/reference/configuration/index.html
```diff
@@ -152,6 +152,13 @@
   <p>
     The <sky-code>cssPath</sky-code> configuration option specifies a path to reference CSS styles. This property is specific to the SKY UX docs site and is for internal Blackbaud use only. To bundle CSS or SCSS files with your SPA, use the <sky-code>styles</sky-code> property of <a stacheRouterLink="." fragment="app">the <sky-code>app</sky-code> configuration option</a> instead.
   </p>
+  
+  <stache-page-anchor>
+    <sky-code>enableIvy</sky-code>
+  </stache-page-anchor>
+  <p>
+    This <sky-code>enableIvy</sky-code> configuration option indicates whether to run Angular Ivy alongside AOT compilation by default. <a href="https://angular.io/guide/ivy">Angular Ivy</a> reduces compile time and bundle size during <sky-code>skyux build</sky-code>. To disable this configuration during builds, set this property to <sky-code>false</sky-code>.
```
blackbaud-johnly, line 160: Completely trivial, but every other configuration property description starts with `The` instead of `This`, so it's probably worth changing this for consistency. ... The only other thing I see is that it might be worth making the first instance of `Angular Ivy` the link instead of the second instance. But again, superminor, so no big deal if you don't want to make the change.

PR #1014, opened 2020-09-25
doc: clarify plugin-resources

Blackbaud-ColbyWhite:
the previous wording implied `/src/app/public/plugin-resources/plugin-resources`, which i'm pretty sure is wrong.

### DIFF
--- a/src/app/learn/get-started/advanced/localization/index.html
+++ b/src/app/learn/get-started/advanced/localization/index.html
@@ -50,7 +50,7 @@
   <ol>
     <li>
       <p>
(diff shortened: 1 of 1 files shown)


blackbaud-johnly:
@Blackbaud-ColbyWhite, I approved but then realized that this change is in a fork, so checks won't pass. I have another PR at https://github.com/blackbaud/skyux2-docs/pull/1016 making changes to the same file, so I'm going to pull your change into that PR and close this one in preference of that to get the checks to pass. Thanks for passing this along, and I'll make sure Steve confirms the change. (I assume it's right because it doesn't make sense to have the duplicate in the path, but I'll double-check with him just to make sure.)

